Issue #11, opened 2016-08-29
Stack buffers leading to stack overflow


c1728p9:
The main stack for most platforms is 4K. The tls-client example uses ~6K of stack which causes a stack overflow. TLS itself seems to be using nearly 4k and main.cpp is using ~2k.
Some places in main.cpp contributing to this stack usage.
https://github.com/ARMmbed/mbed-os-example-tls/blob/master/tls-client/main.cpp#L270
https://github.com/ARMmbed/mbed-os-example-tls/blob/master/tls-client/main.cpp#L373
https://github.com/ARMmbed/mbed-os-example-tls/blob/master/tls-client/main.cpp#L437
https://github.com/ARMmbed/mbed-os-example-tls/blob/master/tls-client/main.cpp#L470

This is causing the issue reported here:
https://github.com/ARMmbed/mbed-os/issues/2567

c1728p9:
@pjbakker, what kind of stack usage would you expect from tls? Are there config options to reduce this?

c1728p9:
I created #12 to reduce some of the stack usage by addressing the above points.

pjbakker:
@c1728p9 Anywhere from 40k-256k depending on configuration and running ciphersuites / certificates?
There are a lot of config options available in *config.h*

This [article on reducing mbed TLS footprint](https://tls.mbed.org/kb/how-to/reduce-mbedtls-memory-and-storage-footprint) might be a good start?

c1728p9:
Hi @pjbakker, are you referring to the stack or the heap? There are many boards with a heap >40K. In mbed-os the main thread's stack is no bigger than 4K for any cortex-m board.

pjbakker:
@c1728p9 Sorry.. Was referring to ROM size. Misread your question.
Stack size depends even more on the called function(tree).
Can you elaborate a bit on the functions stacked?

c1728p9:
I'm seeing the heavy stack usage on mbed-os-example-tls. I have links above to certain functions which are declaring large objects on the stack. Also inside tls there is heavy stack usage, although I don't have precise locations. I wasn't sure if you had any upper bound for stack usage of TLS as you do for the heap usage.

pjbakker:
No, I don't think so.

yanesca:
I have some ideas to reduce stack size, but I don't know if we want to do any of these:

[...]

PR #12, opened 2016-08-29
Reduce example stack usage

c1728p9:
Dynamically allocate buffers that would normally be declared on the
stack. This reduces the maximum stack depth by ~2k.

Note - this just reduces stack usage, the main stack can still overflow even with this patch.

### DIFF
--- a/tls-client/main.cpp
+++ b/tls-client/main.cpp
@@ -267,8 +267,9 @@ public:
         /* It also means the handshake is done, time to print info */
         printf("TLS connection to %s established\r\n", HTTPS_SERVER_NAME);
 
-        char buf[1024];
-        mbedtls_x509_crt_info(buf, sizeof(buf), "\r    ",
+        const uint32_t buf_size = 1024;
+        char *buf = new char[buf_size];
+        mbedtls_x509_crt_info(buf, buf_size, "\r    ",
                         mbedtls_ssl_get_peer_cert(&_ssl));
         mbedtls_printf("Server certificate:\r\n%s\r", buf);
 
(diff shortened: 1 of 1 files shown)


adustm:
This PR is needed to pass tls-client on NUCLEO_F429ZI.

tls-client/main.cpp
```diff
@@ -310,6 +312,7 @@ class HelloHTTPS {
         _error = !(_got200 && _gothello);
 
         _tcpsocket->close();
+        delete buf;
```
sbutcher-arm: Second call to `delete buf;` which is a double free.

sbutcher-arm:
Pull request is to fix #11.

sbutcher-arm:
Reviewed and +1 from me. 

@yanesca - Could you please test the PR on a board, and then merge it?

tls-client/main.cpp
```diff
@@ -291,6 +292,7 @@ class HelloHTTPS {
                 print_mbedtls_error("mbedtls_ssl_read", ret);
                 onError(_tcpsocket, -1 );
             }
+            delete buf;
```
andresag01: Shouldnt this be `delete[] buf`?

tls-client/main.cpp
andresag01: Shouldnt this be `delete[] buf;`?

c1728p9:
Thanks for the review @andresag01. I updated the PR to address your comments.

yanesca:
Reviewed, applied @andresag01 's suggestions, tested and merged.

PR #17, opened 2016-09-30
Enable SHA1 for tls-client and remove old cert

andresag01:


andresag01:
I have marked it as draft because the trusted CA for tls-client needs to be replaced.

andresag01:
Closing because the PR is moved to oob repo https://github.com/ARMmbed/oob-mbed-os-example-tls

PR #18, opened 2016-10-03
Fix benchmark entropy_config and add license

andresag01:


andresag01:
Closing because the PR is moved to oob repo https://github.com/ARMmbed/oob-mbed-os-example-tls

PR #20, opened 2016-10-27
Fix mbedtls_entropy_config.h condition

andresag01:
This PR introduces two changes:
- Remove unnecessary metadata files created by mbed-cli that should not be committed with the source code.
- Tighten condition of mbedtls_entropy_config.h to ensure that the mbedtls entropy module is enabled when compiling authcrypt and tls-client examples.


andresag01:
@yanesca @sbutcher-arm: please review.

yanesca:
I don't think we should make these changes:
- If you build/update an example without those metadata files the mbed tool won't stop after building/updating the example, it will partially build/update all the other examples as well.
- Here the assumption is that the configuration is the default mbed-OS configuration, and the check in the `mbedtls_entropy_config.h` is there exclusively for testing device features, therefore I don't feel this extension of the condition necessary.

andresag01:
Closing PR as suggested

PR #30, opened 2016-12-05
Campare logs for verification

mazimkhan:
Compare logs for example verification in CI. Depends on https://github.com/ARMmbed/htrun/pull/139
See format details here https://github.com/mazimkhan/htrun/blob/a6e5a04a60aecab408dfd2f6d62939cb69e3ae1c/README.md#testing-mbed-os-examples


mazimkhan:
@sbutcher-arm @andresag01 Please review.

mazimkhan:
@sbutcher-arm @andresag01 
- moved logs to ```tests``` directory.
- Added missing ```DONE``` in ```tests/benchmark.log```
- Created ```tests/README.md```

tests/README.md
```diff
@@ -0,0 +1,156 @@
...
+							      >	Strict-Transport-Security: max-age=31536000; includeSubdomain
+								
+
+								
+								
+Hello world!								Hello world!
+
+```
+
+Please observe above that all the lines that have data that changes from execution to execution (on right) have been removed. It makes it possible htrun to compare these logs. htrun matches lines from the compare log (on left) one by one. It keeps on looking for a line until it matches. Once matched it moves on to match the next line. If it finds all lines from the compare log in the target serial output stream. Then it halts and passes the examples.
```
andresag01: I suggest: `Then it halts and passes the examples.` -> `Then it halts and the test passes.`

tests/README.md
```diff
@@ -0,0 +1,156 @@
...
+							      >	X-Upstream-L2: developer-sjc-indigo-2-nginx
+								
+							      >	Strict-Transport-Security: max-age=31536000; includeSubdomain
+								
+
+								
+								
+Hello world!								Hello world!
+
+```
```
andresag01: I think the rationale of the `README` is to provide some insight regarding the `.log` files. I think that having a short example is appropriate, however, I consider that the examples of htrun usage in this README are a bit too extensive and better suited to the htrun documentation itself. I suggest removing the tls-client example and simply leaving a fragment of the benchmark log comparison below (perhaps 4-5 lines). What are your thoughts @mazimkhan?

tests/authcrypt.log
```diff
@@ -0,0 +1,4 @@
+plaintext message: [0-9a-fA-F]*
+ciphertext: [0-9a-fA-F]*
+decrypted: [0-9a-fA-F]*
```
andresag01: I think this regular expression is a little too simple. This will match something like:
```
plaintext message: 
ciphertext: 
decrypted: 
```
Which is not correct output because there is no hash value. Also, we know that the example uses SHA256 so the output length is always the same. Is it possible to change the expression to match a fixed number of characters? Also, SHA256 does not have any randomness, I do not think the output is expected to change between runs, but perhaps you should check that, because then you do not really need a regexp but a fixed value.

tests/README.md
mazimkhan: I would leave this example as it illustrates the one way of making a template log that is by removing random data.

tests/authcrypt.log
mazimkhan: Changed ```[0-9a-fA-F]*``` -> ```[0-9a-fA-F]+```.

andresag01: Sorry, my previous message was a bit misleading because I thought this was hashing instead of authcrypt. I still think bits of it do not need a complex regexp. For example the plaintext message and the decrypted message must be the same value and is hardcoded in the example.

tests/authcrypt.log
```diff
@@ -0,0 +1,4 @@
+plaintext message: [0-9a-fA-F]*
+ciphertext: [0-9a-fA-F]*
+decrypted: [0-9a-fA-F]*
+DONE
```
mazimkhan: These are because the template is made out of a serial output. It is not hand made. We can remove ```\r``` for readability but that will require manual effort of making the logs readable.

tests/benchmark.log
```diff
@@ -0,0 +1,26 @@
...

+  ECDHE-secp256r1          :\s*\d+ ms/handshake

+  ECDHE-Curve25519         :\s*\d+ ms/handshake

+  ECDH-secp384r1           :\s*\d+ ms/handshake

+  ECDH-secp256r1           :\s*\d+ ms/handshake

+  ECDH-Curve25519          :\s*\d+ ms/handshake
```
mazimkhan: We can do that if the code is suppose to change this space. Since it is fixed space we don't need to worry about it. Idea is to keep this approach simple and stupid and only put effort in regex for complex patterns.

Also one thing to bear in mind is that htrun first tries to compare a line character by character. In this case a line with special characters will also match. But if this comparison fails then htrun tries to use the line form the template as a regex. Hence if we remove any character from the template line (i.e. treating it as a regex) then we have to be careful of the characters in the line (regex). As characters like ```()<>.*+\``` are used as special meaning chars in the regex. They would then need to be escaped.

mazimkhan:
@andresag01 Incorporated code review comments. Please review.

andresag01:
@Azim: As discussed before, could you please remove the `\r` from the pattern files? Also, it seems that there are some issues with `htrun` in Linux.

mazimkhan:
@andresag01 good find. Removed carriage return from the logs and it works for me now. Please try again.

andresag01:
@mazimkhan: Thanks for fixing the issues previously found. However, I am having difficulties trying to test these log files as `mbedhtrun` is not able to properly flash applications onto the target reliably in Linux. I get the following error:
```
[1485874580.84][COPY][INF] Waiting for mount point '/media/andama01/MBED' to be ready...
..............................cp: cannot create regular file '/media/andama01/MBED/benchmark.bin': No such file or directory
[1485874588.56][COPY][ERR] [ret=1] Command: ['cp', 'BUILD/K64F/GCC_ARM/benchmark.bin', '/media/andama01/MBED/benchmark.bin']
```
Furthermore, if I manually mount the device and then run `mbedhtrun` then the program works in some cases but it fails in some other cases with the following error:
```
[1485874744.02][HTST][INF] test suite run finished after 60.09 sec...
[1485874744.02][HTST][ERR] Expected output [DONE] not received in log.
[1485874744.03][CONN][INF] received special even '__host_test_finished' value='True', finishing
[1485874744.03][HTST][INF] CONN exited with code: 0
[1485874744.03][HTST][INF] Some events in queue
[1485874744.03][HTST][INF] stopped consuming events
[1485874744.03][HTST][INF] host test result(): None
[1485874744.03][HTST][WRN] missing __exit event from DUT
[1485874744.03][HTST][WRN] missing __exit_event_queue event from host test
[1485874744.03][HTST][ERR] missing __exit_event_queue event from host test and no result from host test, timeout...
[1485874744.03][HTST][INF] calling blocking teardown()
[1485874744.03][HTST][INF] teardown() finished
[1485874744.03][HTST][INF] {{result;timeout}}
```

mazimkhan:
I believe you have some mounting issues. Can you also supply ‘-P 120’ option. 120 means 2 minutes. –P increases the timeout and may give enough time to htrun to remount the mbed device and run the test.

andresag01:
I ran the test for each of the examples and it seems to work as expected.

PR #36, opened 2016-12-21
Update mbed-os.lib to mbed-os-5.3.1

adbridge:


andresag01:
I have manually tested the changes with GCC_ARM and ARM in K64F and it seems work correctly.

PR #40, opened 2017-01-24
Conditionally def MBEDTLS_SHA1_C to avoid redef

andresag01:
The tls-client example requires SHA1 because it is needed to be ale to verify certificates. However, this feature is disabled by default in mbed OS, so the user-defined config.h defines the macro MBEDTLS_SHA1_C to enable it. However, some platforms have drivers that also define MBEDTLS_SHA1_C, which might result in several macro redefinition warning during compilation. This patch solves the problem by defining MBEDTLS_SHA1_C in the tls-client user-config only if it is not already defined.


yanesca:
Looks good to me!

PR #41, opened 2017-01-25
Disable EMAC feature in ublox to fix tls-client

andresag01:
When the macro DEVICE_EMAC is enabled, the program will try to use Wifi
by default. However, the tls-client example only supports Ethernet, so
it fails. To avoid this we use the device_has_remove field in the
mbed_app.json to disable the EMAC feature for UBLOX_EVK_ODIN_W2.


yanesca:
Would this problem arise with every board that has DEVICE_EMAC is enabled? If so, then is there a way to solve this issue for every board with wifi in general and not just for UBLOX_EVK_ODIN_W2?

PR #49, opened 2017-02-16
DRAFT: Gracefully close TLS connection in tls-client

andresag01:
Gracefully terminate the TLS connection to the remote server once all
data has been successfully received. This change also adds some error
checks.

PR #51, opened 2017-02-28
Fix tls client socket io - DRAFT

yanesca:
This branch is based on PR #49

### DIFF
--- a/tls-client/main.cpp
+++ b/tls-client/main.cpp
@@ -469,8 +469,7 @@ int main() {
     hello->startTest(HTTPS_PATH);
     if (hello->error()) {
         mbedtls_printf("FAIL\r\n");
-    }
-    else {
+    } else {
         mbedtls_printf("DONE\r\n");
     }
     delete hello;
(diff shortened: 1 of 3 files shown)


tls-client/main.cpp
```diff
@@ -196,108 +216,118 @@ class HelloHTTPS {
...
+        _gothello = _gothello || strstr(_recv_buf, HTTPS_HELLO_STR) != NULL;
 
         /* Print status messages */
         mbedtls_printf("HTTPS: Received %d chars from server\r\n", _bpos);
         mbedtls_printf("HTTPS: Received 200 OK status ... %s\r\n", _got200 ? "[OK]" : "[FAIL]");
         mbedtls_printf("HTTPS: Received '%s' status ... %s\r\n", HTTPS_HELLO_STR, _gothello ? "[OK]" : "[FAIL]");
         mbedtls_printf("HTTPS: Received message:\r\n\r\n");
-        mbedtls_printf("%s", _buffer);
+        mbedtls_printf("%s", _recv_buf);
         _error = !(_got200 && _gothello);
```
mazimkhan: _error should be set at all places where we jump to exit. Else it will always print DONE.

PR #53, opened 2017-03-01
Fix handling MBEDTLS_ERR_SSL_WANT_READ with mbedtls functions

mazimkhan:
mbedtls functions that may return MBEDTLS_ERR_SSL_WANT_WRITE or MBEDTLS_ERR_SSL_WANT_READ should be called again until there is an error or they succeed.

### DIFF
--- a/tls-client/main.cpp
+++ b/tls-client/main.cpp
@@ -55,7 +55,6 @@ const int HTTPS_SERVER_PORT = 443;
 const int RECV_BUFFER_SIZE = 600;
 
 const char HTTPS_PATH[] = "/media/uploads/mbed_official/hello.txt";
-const size_t HTTPS_PATH_LEN = sizeof(HTTPS_PATH) - 1;
 
 /* Test related data */
 const char *HTTPS_OK_STR = "200 OK";
@@ -110,12 +109,13 @@ public:
             _domain(domain), _port(port)
     {
 
-        _error = false;
         _gothello = false;
         _got200 = false;
         _bpos = 0;
         _request_sent = 0;
         _tcpsocket = new TCPSocket(net_iface);
+        _tcpsocket->set_blocking(false);
+        _buffer[RECV_BUFFER_SIZE - 1] = 0;
 
         mbedtls_entropy_init(&_entropy);
         mbedtls_ctr_drbg_init(&_ctr_drbg);
@@ -132,6 +132,8 @@ public:
         mbedtls_x509_crt_free(&_cacert);
         mbedtls_ssl_free(&_ssl);
         mbedtls_ssl_config_free(&_ssl_conf);
+        _tcpsocket->close();
+        delete _tcpsocket;
     }
     /**
      * Start the test.
@@ -145,11 +147,8 @@ public:
         /* Initialize the flags */
         _got200 = false;
         _gothello = false;
-        _error = false;
         _disconnected = false;
         _request_sent = false;
-        /* Fill the request buffer */
-        _bpos = snprintf(_buffer, sizeof(_buffer) - 1, "GET %s HTTP/1.1\nHost: %s\n\n", path, HTTPS_SERVER_NAME);
 
         /*
          * Initialize TLS-related stuf.
@@ -159,14 +158,12 @@ public:
                           (const unsigned char *) DRBG_PERS,
                           sizeof (DRBG_PERS))) != 0) {
             print_mbedtls_error("mbedtls_crt_drbg_init", ret);
-            _error = true;
             return;
         }
 
         if ((ret = mbedtls_x509_crt_parse(&_cacert, (const unsigned char *) SSL_CA_PEM,
                            sizeof (SSL_CA_PEM))) != 0) {
             print_mbedtls_error("mbedtls_x509_crt_parse", ret);
-            _error = true;
             return;
         }
 
@@ -175,7 +172,6 @@ public:
                         MBEDTLS_SSL_TRANSPORT_STREAM,
                         MBEDTLS_SSL_PRESET_DEFAULT)) != 0) {
             print_mbedtls_error("mbedtls_ssl_config_defaults", ret);
-            _error = true;
             return;
         }
 
@@ -195,7 +191,6 @@ public:
 
         if ((ret = mbedtls_ssl_setup(&_ssl, &_ssl_conf)) != 0) {
             print_mbedtls_error("mbedtls_ssl_setup", ret);
-            _error = true;
             return;
         }
 
@@ -210,29 +205,39 @@ public:
         ret = _tcpsocket->connect(_domain, _port);
         if (ret != NSAPI_ERROR_OK) {
             mbedtls_printf("Failed to connect\r\n");
-            onError(_tcpsocket, -1);
+            printf("MBED: Socket Error: %d\r\n", ret);
+            _tcpsocket->close();
             return;
         }
 
        /* Start the handshake, the rest will be done in onReceive() */
         mbedtls_printf("Starting the TLS handshake...\r\n");
-        ret = mbedtls_ssl_handshake(&_ssl);
+        do {
+            ret = mbedtls_ssl_handshake(&_ssl);
+        } while (ret != 0 && (ret == MBEDTLS_ERR_SSL_WANT_READ ||
+                ret == MBEDTLS_ERR_SSL_WANT_WRITE));
         if (ret < 0) {
-            if (ret != MBEDTLS_ERR_SSL_WANT_READ &&
-                ret != MBEDTLS_ERR_SSL_WANT_WRITE) {
-                print_mbedtls_error("mbedtls_ssl_handshake", ret);
-                onError(_tcpsocket, -1 );
-            }
+            print_mbedtls_error("mbedtls_ssl_handshake", ret);
+            _tcpsocket->close();
             return;
         }
 
-        ret = mbedtls_ssl_write(&_ssl, (const unsigned char *) _buffer, _bpos);
+        /* Fill the request buffer */
+        _bpos = snprintf(_buffer, sizeof(_buffer) - 1, 
+                         "GET %s HTTP/1.1\nHost: %s\n\n", path, HTTPS_SERVER_NAME);
+
+        int offset = 0;
+        do {
+            ret = mbedtls_ssl_write(&_ssl, 
+                                    (const unsigned char *) _buffer + offset, 
+                                    _bpos - offset);
+            if (ret > 0)
+              offset += ret;
+        } while (offset < _bpos && (ret > 0 || ret == MBEDTLS_ERR_SSL_WANT_READ ||
+                ret == MBEDTLS_ERR_SSL_WANT_WRITE));
         if (ret < 0) {
-            if (ret != MBEDTLS_ERR_SSL_WANT_READ &&
-                ret != MBEDTLS_ERR_SSL_WANT_WRITE) {
-                print_mbedtls_error("mbedtls_ssl_write", ret);
-                onError(_tcpsocket, -1 );
-            }
+            print_mbedtls_error("mbedtls_ssl_write", ret);
+            _tcpsocket->close();
             return;
         }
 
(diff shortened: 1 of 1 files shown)


mazimkhan:
@yanesca Please review.

tls-client/main.cpp
```diff
@@ -132,6 +133,8 @@ class HelloHTTPS {
         mbedtls_x509_crt_free(&_cacert);
         mbedtls_ssl_free(&_ssl);
         mbedtls_ssl_config_free(&_ssl_conf);
+        _tcpsocket->close();
```
mazimkhan: ```close()``` is a harmless method and if you see its implementation, it checks if the socket is open or not. Moving it to the destructor add a lot of simplicity as we don't have to call it every time we ```return``` from ```startTest``` or use ```goto``` statements.

tls-client/main.cpp
```diff
@@ -216,23 +219,21 @@ class HelloHTTPS {
...
+            ret = mbedtls_ssl_handshake(&_ssl);
+        } while (ret != 0 && (ret == MBEDTLS_ERR_SSL_WANT_READ ||
+                ret == MBEDTLS_ERR_SSL_WANT_WRITE));
         if (ret < 0) {
-            if (ret != MBEDTLS_ERR_SSL_WANT_READ &&
-                ret != MBEDTLS_ERR_SSL_WANT_WRITE) {
-                print_mbedtls_error("mbedtls_ssl_handshake", ret);
-                onError(_tcpsocket, -1 );
-            }
+            print_mbedtls_error("mbedtls_ssl_handshake", ret);
```
mazimkhan: Since ```close``` is moved to the destructor we don't need call to ```onError```. Major reason of not calling ```onError``` is that it misleads in the logs that error has occurred due to Socket.

mazimkhan: It applies to all comments for same reason below.

tls-client/main.cpp
```diff
@@ -301,7 +301,7 @@ class HelloHTTPS {
     /**
      * Closes the TCP socket
      */
-    void close() {
+    void close() { // Do we need this functions?
```
mazimkhan: Agreed and removed.

tls-client/main.cpp
yanesca: My main reason for suggesting this change wasn't some effective gain on the robustness of the application, I merely prefer that solution because of the logical coherence of the code.

tls-client/main.cpp
yanesca: If we leave closing the socket to the destructor, then we have a different problem: the `onError()` function becomes obsolete and we need to remove it and replace each call with the line `_error = true;`.

mazimkhan: member ```_error```, functions ```error()``` and ```done``` aren't used anywhere. Hence setting it also doesn't make sense. looks more like incomplete code. I would rather remove it.

tls-client/main.cpp
mazimkhan: I see more lines of code to manage closing the socket in ```startTest()``` than putting it in the constructor. We have to see trade off between coherence and pollution in code. For coherence  we should also return from a single point in ```startTest```.

yanesca: Ok, then let's do that: move the code that opens the connection to the constructor!

tls-client/main.cpp
yanesca: Great! Then let's remove them!

mazimkhan:
@yanesca Incorporated your comments and also made some improvements for mbedtls_ssl_read/write calls.

tls-client/main.cpp
```diff
@@ -256,22 +261,31 @@ class HelloHTTPS {
...
+        int read = 0;
         do {
-            ret = mbedtls_ssl_read(&_ssl, (unsigned char *) _buffer, sizeof(_buffer));
-        } while (ret != 0 && (ret == MBEDTLS_ERR_SSL_WANT_READ ||
-                ret == MBEDTLS_ERR_SSL_WANT_WRITE));
+            ret = mbedtls_ssl_read(&_ssl, (unsigned char *) _buffer + read, 
+                                   sizeof(_buffer) - 1);
+            if (ret > 0)
+              read += ret;
+            else if (ret != MBEDTLS_ERR_SSL_WANT_READ &&
```
yanesca: Why did you extract this condition from the head of the loop?

tls-client/main.cpp
```diff
@@ -256,22 +261,31 @@ class HelloHTTPS {
...
         }
-        _bpos = static_cast<size_t>(ret);
+        _bpos = static_cast<size_t>(read);
 
         _buffer[_bpos] = 0;
-
-        /* Check each of the flags */
-        _got200 = _got200 || strstr(_buffer, HTTPS_OK_STR) != NULL;
-        _gothello = _gothello || strstr(_buffer, HTTPS_HELLO_STR) != NULL;
+
```
yanesca: There are whitespace characters on this empty line, could you please remove them?

tls-client/main.cpp
```diff
@@ -401,12 +412,11 @@ class HelloHTTPS {
     void onError(TCPSocket *s, int error) {
         printf("MBED: Socket Error: %d\r\n", error);
         s->close();
-        _error = true;
     }
 
 protected:
     TCPSocket* _tcpsocket;
-
+
```
yanesca: There are whitespace characters on this empty line, could you please remove them?

tls-client/main.cpp
```diff
@@ -148,9 +151,7 @@ class HelloHTTPS {
         _error = false;
         _disconnected = false;
         _request_sent = false;
-        /* Fill the request buffer */
-        _bpos = snprintf(_buffer, sizeof(_buffer) - 1, "GET %s HTTP/1.1\nHost: %s\n\n", path, HTTPS_SERVER_NAME);
-
+
```
yanesca: There are whitespace characters on this empty line, could you please remove them?

tls-client/main.cpp
```diff
@@ -216,23 +213,31 @@ class HelloHTTPS {
...
-            }
+            print_mbedtls_error("mbedtls_ssl_handshake", ret);
+            _tcpsocket->close();
             return;
         }
 
-        ret = mbedtls_ssl_write(&_ssl, (const unsigned char *) _buffer, _bpos);
+        /* Fill the request buffer */
+        _bpos = snprintf(_buffer, sizeof(_buffer) - 1, "GET %s HTTP/1.1\nHost: %s\n\n", path, HTTPS_SERVER_NAME);
+
```
yanesca: There are whitespace characters on this empty line, could you please remove them?

tls-client/main.cpp
```diff
@@ -401,12 +412,11 @@ class HelloHTTPS {
     void onError(TCPSocket *s, int error) {
```
yanesca: This function has only a single call on line 210 and the string it prints conveys no additional information. I think the call on line 210 can be replaced with `_tcpsocket->close();` and this function can be deleted.

tls-client/main.cpp
```diff
@@ -224,15 +219,25 @@ class HelloHTTPS {
...
+            _tcpsocket->close();
             return;
         }
 
+        /* Fill the request buffer */
+        _bpos = snprintf(_buffer, sizeof(_buffer) - 1, "GET %s HTTP/1.1\nHost: %s\n\n", path, HTTPS_SERVER_NAME);
+        
+        int written = 0;
         do {
             ret = mbedtls_ssl_write(&_ssl, (const unsigned char *) _buffer, _bpos);
```
yanesca: The [manual for this function](https://tls.mbed.org/api/ssl_8h.html) says that it has to be called with updated arguments in case the write was partial.

tls-client/main.cpp
```diff
@@ -256,22 +261,31 @@ class HelloHTTPS {
...
+                                   sizeof(_buffer) - 1);
+            if (ret > 0)
+              read += ret;
+            else if (ret != MBEDTLS_ERR_SSL_WANT_READ &&
+                ret != MBEDTLS_ERR_SSL_WANT_WRITE)
+              break;
+            /* Check each of the flags */
+            _got200 = _got200 || strstr(_buffer, HTTPS_OK_STR) != NULL;
+            _gothello = _gothello || strstr(_buffer, HTTPS_HELLO_STR) != NULL;
+        } while (!_got200 || !_gothello);
```
mazimkhan: Because we don't know how much to read before checking for these strings or how long to run the read loop.

tls-client/main.cpp
mazimkhan: Because this condition is not sufficient as the loop condition. ```ret``` can have positive values. But I can squeeze the new condition in while if you want.

tls-client/main.cpp
```diff
@@ -224,15 +219,25 @@ class HelloHTTPS {
...
+        _bpos = snprintf(_buffer, sizeof(_buffer) - 1, "GET %s HTTP/1.1\nHost: %s\n\n", path, HTTPS_SERVER_NAME);
+        
+        int written = 0;
         do {
             ret = mbedtls_ssl_write(&_ssl, (const unsigned char *) _buffer, _bpos);
-        } while (ret != 0 && (ret == MBEDTLS_ERR_SSL_WANT_READ ||
-                ret == MBEDTLS_ERR_SSL_WANT_WRITE));
+            if (ret > 0)
+              written += ret;
+            else if (ret != MBEDTLS_ERR_SSL_WANT_READ &&
```
mazimkhan: Because this condition is not sufficient as the loop condition. ```ret``` can have positive values. But I can squeeze the new condition in while if you want.

tls-client/main.cpp
mazimkhan: Updated

tls-client/main.cpp
```diff
@@ -256,22 +263,32 @@ class HelloHTTPS {
...
+            ret = mbedtls_ssl_read(&_ssl, (unsigned char *) _buffer + offset, 
+                                   sizeof(_buffer) - offset - 1);
+            if (ret > 0)
+              offset += ret;
+
+            /* Check each of the flags */
+            _buffer[offset] = 0;
+            _got200 = _got200 || strstr(_buffer, HTTPS_OK_STR) != NULL;
+            _gothello = _gothello || strstr(_buffer, HTTPS_HELLO_STR) != NULL;
+        } while (offset < sizeof(_buffer) && (!_got200 || !_gothello) &&
```
yanesca: Here we can omit the `offset < sizeof(_buffer)` condition, because `mbedtls_ssl_read` should check it and throw an error if the data does not fit.

tls-client/main.cpp
```diff
@@ -256,22 +263,32 @@ class HelloHTTPS {
...
+                (ret || ret == MBEDTLS_ERR_SSL_WANT_READ ||
```
yanesca: I know that the functionality is the same, but would you mind if we would write (ret > 0 instead of (ret here for better readability?

tls-client/main.cpp
yanesca: It is removed in the latest version.

tls-client/main.cpp
yanesca: In the latest version the constructor remained the same and the function closes the connection before exiting and this approach resolves my above concern as well.

yanesca:
Looks good to me!

mazimkhan:
@andresag01 @sbutcher-arm Please review and merge.

tls-client/main.cpp
```diff
@@ -256,22 +263,32 @@ class HelloHTTPS {
...
+            if (ret > 0)
+              offset += ret;
+
+            /* Check each of the flags */
+            _buffer[offset] = 0;
+            _got200 = _got200 || strstr(_buffer, HTTPS_OK_STR) != NULL;
+            _gothello = _gothello || strstr(_buffer, HTTPS_HELLO_STR) != NULL;
+        } while ( (!_got200 || !_gothello) &&
+                (ret > 0 || ret == MBEDTLS_ERR_SSL_WANT_READ ||
+                ret == MBEDTLS_ERR_SSL_WANT_WRITE));
```
andresag01: Why are these loops here instead of the `ssl_send()` and `ssl_recv()` functions? It seems that we are having loops over and over. Perhaps it is better to just put them in one place?

tls-client/main.cpp
```diff
@@ -210,29 +207,39 @@ class HelloHTTPS {
         ret = _tcpsocket->connect(_domain, _port);
         if (ret != NSAPI_ERROR_OK) {
             mbedtls_printf("Failed to connect\r\n");
-            onError(_tcpsocket, -1);
+            printf("MBED: Socket Error: %d\r\n", ret);
+            _tcpsocket->close();
```
mazimkhan: For the first comment. Having a status print in ```main()``` deviates from the issue this PR addresses. Secondly that would make this example look more like a test case. Success or failure is apparent by looking the progress in the logs.

Second comment: putting ```_tcpsocket->close()``` in destructor was a good idea. But @yanesca wanted to maintain coherence by also moving ```_tcpsocket->connect()``` in the constructor. That I don't agree with that as connection failure will result in an object that isn't in the expected state.

mazimkhan: Hence instead I called it at each exit point. This applies to all similar comments.

tls-client/main.cpp
mazimkhan: Since mbedtls lib expects and handles non blocking IO. The ssl_send and recv routines should not be blocking in terms of looping until data is received. Secondly in some cases like call to ```mbedtls_ssl_read```  the application does not know how much it want to receive and it sends buffer size instead of expected message size. Hence putting such loops in those routines can cause app to hang.

tls-client/main.cpp
```diff
@@ -375,6 +387,7 @@ class HelloHTTPS {
         if(NSAPI_ERROR_WOULD_BLOCK == recv){
             return MBEDTLS_ERR_SSL_WANT_READ;
         }else if(recv < 0){
+            mbedtls_printf("Socket recv error %d\r\n", recv);
             return -1;
```
mazimkhan: Did not intend to change this. This could be pat of a different PR after analysing why it was done like this originally.

tls-client/main.cpp
```diff
@@ -390,20 +403,15 @@ class HelloHTTPS {
         size = socket->send(buf, len);
 
         if(NSAPI_ERROR_WOULD_BLOCK == size){
-            return len;
+            return MBEDTLS_ERR_SSL_WANT_WRITE;
         }else if(size < 0){
+            mbedtls_printf("Socket send error %d\r\n", size);
             return -1;
```
mazimkhan: same as above.

mazimkhan:
@andresag01 cleaned up unused ```_error``` and unused methods. Please reply to the comments.

andresag01:
I have manually merged these changes into development branch. Refer to https://github.com/ARMmbed/mbed-os-example-tls/commit/4459c46e976ccf754f75e0028d8eb2a4041a2463

mazimkhan:
@andresag01 @sbutcher-arm why is this not merged into master?

andresag01:
@mazimkhan we are planning to do some additional work on this example. It was decided that we will continue in a development branch and then integrate into master once we have decided the example is stable.

mazimkhan:
Ok but don't understand this "once we have decided the example is stable."

My understanding was that changes planned in the development branch were for minor refactoring of this example. The example has been made stable in this PR i.e. 53. The refactoring shouldn't be related or prompted by this PR i.e. could have been done on the code before.

PR #59, opened 2017-03-04
Update mbed-os.lib to fix memory issue in tls-client test on UBLOX IA…

mazimkhan:
…R build


andresag01:
This PR updates the `mbed-os.lib` file to an mbed OS commit that is not tagged with a release, which is not common practice in the example code. I will close this PR as the relevant commit from mbed OS will be pulled into the example when we update the `mbed-os.lib` for the next mbed OS release.

PR #63, opened 2017-03-08
Refactor tls-client example to improve readability

andresag01:
This patch completely refactors the tls-client example. The following
is a list of significant changes:
    * Split the code into .h and .cpp files.
    * Move the class HelloHTTPS to its own file.
    * Remove all unused variables and redundant buffers.
    * Split functionality into smaller, more manageable functions.
    * Improve error reporting and logging.

### DIFF
--- a/tls-client/HelloHttpsClient.cpp
+++ b/tls-client/HelloHttpsClient.cpp
@@ -1,0 +1,431 @@
+/*
+ *  Hello world example of a TLS client: fetch an HTTPS page
+ *
+ *  Copyright (C) 2006-2017, ARM Limited, All Rights Reserved
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *  not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  This file is part of mbed TLS (https://tls.mbed.org)
+ */
+
+#include "HelloHttpsClient.h"
+
+#include "NetworkInterface.h"
+#include "TCPSocket.h"
+
+#include "mbedtls/platform.h"
+#include "mbedtls/config.h"
+#include "mbedtls/ssl.h"
+#include "mbedtls/entropy.h"
+#include "mbedtls/ctr_drbg.h"
+#include "mbedtls/error.h"
+#include "mbedtls/debug.h"
+
+#include <stdint.h>
+#include <string.h>
+
+HelloHttpsClient::HelloHttpsClient( const char *in_server_name,
+                                    const uint16_t in_server_port ) :
+    eth_iface( NULL ),
+    socket( NULL ),
+    server_name( in_server_name ),
+    server_port( in_server_port )
+{
+    mbedtls_entropy_init(&entropy);
+    mbedtls_ctr_drbg_init(&ctr_drbg);
+    mbedtls_x509_crt_init(&cacert);
+    mbedtls_ssl_init(&ssl);
+    mbedtls_ssl_config_init(&ssl_conf);
+}
+
+HelloHttpsClient::~HelloHttpsClient( void )
+{
+    mbedtls_entropy_free(&entropy);
+    mbedtls_ctr_drbg_free(&ctr_drbg);
+    mbedtls_x509_crt_free(&cacert);
+    mbedtls_ssl_free(&ssl);
+    mbedtls_ssl_config_free(&ssl_conf);
+
+    if( eth_iface != NULL )
+        delete eth_iface;
+    if( socket != NULL )
+        delete socket;
+}
+
+int HelloHttpsClient::createTCPSocket( void )
+{
+    int ret;
+    const char *ip_addr;
+
+    eth_iface = new EthernetInterface();
+    if( eth_iface == NULL )
+    {
+        mbedtls_printf( "Failed to allocate EthernetInterface\r\n" );
+        return( -1 );
+    }
+
+    /* Initialise the ethernet interface and start up the stack */
+    if( ( ret = eth_iface->connect() ) != 0 )
+    {
+        mbedtls_printf( "Failed call to eth_iface.connect(): %d\r\n", ret );
+        return( ret );
+    }
+
+    ip_addr = eth_iface->get_ip_address();
+    if( ip_addr != NULL )
+    {
+        mbedtls_printf( "Client IP address is %s\r\n", ip_addr );
+    }
+    else
+    {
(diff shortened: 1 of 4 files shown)


andresag01:
@yanesca @mazimkhan @sbutcher-arm: Could you please review?

The CI ran as expected and reported the same known failures:
* NUMAKER_PFM_NUC472 fails with all compilers
* UBLOX_EVK_ODIN_W2 fails with IAR (this fails because the mbed OS patch that fixes it has not yet been released).

tls-client/main.cpp
```diff
@@ -19,417 +19,72 @@
...
-        /* Extract basename from file */
-        for(p = basename = file; *p != '\0'; p++) {
-            if(*p == '/' || *p == '\\') {
-                basename = p + 1;
-            }
-        }
-
-        mbedtls_printf("%s:%04d: |%d| %s", basename, line, level, str);
+        mbedtls_printf( "Failed to allocate HelloHttpsClient object\r\n" );
+        goto exit;
```
mazimkhan: Just for 2 ```if``` blocks I would not use ```goto```. Instead a nested ```if``` would be more clear.

tls-client/main.cpp
```diff
@@ -19,417 +19,72 @@
...
-        if(NSAPI_ERROR_WOULD_BLOCK == size){
-            return MBEDTLS_ERR_SSL_WANT_WRITE;
-        }else if(size < 0){
-            mbedtls_printf("Socket send error %d\r\n", size);
-            return -1;
-        }else{
-            return size;
-        }
+exit:
+    if( exit_code == MBEDTLS_EXIT_SUCCESS )
```
mazimkhan: Why do we have to instrument printing ```DONE``` or ```FAIL```. It looks like a test case output. Is it not clear that an example failed when it does not reach ```DONE```.

tls-client/HelloHttpsClient.h
```diff
@@ -0,0 +1,247 @@
...
+    const char *server_name;
+    /**
+     * The port number to use in the connection
+     */
+    const uint16_t server_port;
+
+    /**
+     * A generic buffer used to hold debug or HTTP request/response strings
+     */
+    char gp_buf[GENERAL_PURPOSE_BUFFER_LENGTH];
```
mazimkhan: No need to pre-allocate this buffer as part of this class. It can be allocated when needed. That is when ssl write is done. Before ssl write the ssl handshake API grows memory for certificate verification and key comparisons. Finally it frees the memory. A pre allocated buffer will keep the mem at that peak time.
It may help it to work on memory constrained platforms. Also a good practice.

tls-client/main.cpp
andresag01: It makes it easier to check from `htrun` whether the example fully worked (including the teardown logic).

tls-client/main.cpp
andresag01: I disagree. Throughout the example and in the mbed TLS library, this is the way we generally handle failures, and in my opinion it is very clear. It also reduces code duplication and helps maintenance as there is only one exit point for the function. I suppose that at the end of the day, this is just favourite coding practices.

tls-client/HelloHttpsClient.h
andresag01: Agreed, assuming that the handshake is the peak consumption point. However, I am not sure this will cause an issue as this is allocated in the heap. What do the other reviewers think?

tls-client/main.cpp
mazimkhan: ```htrun``` has log comparison feature for detecting pass or fail. It does not look for ```DONE``` or ```FAIL```.

tls-client/main.cpp
mazimkhan: It is not about what is already there in lib. ```goto``` is only help when there is functions long sequence of function calls that need different cleanups. Here there are only two. A nested if would be better. But it is just my opinion.

tls-client/main.cpp
andresag01: @mazimkhan I dont understand what you mean. I thought `htrun` only checks the serial output (just as the user would do), how can it tell what the return code of the function is?

tls-client/HelloHttpsClient.cpp
```diff
@@ -0,0 +1,431 @@
...
+    delete[] buf;
+
+exit:
+    return( ret );
+}
+
+int HelloHttpsClient::run( void )
+{
+    int ret;
+    size_t req_len, req_offset, resp_offset;
```
mazimkhan: C++ does not enforce declaring variables at top. They should be as close to the usage as possible. These could declared and assigned at https://github.com/ARMmbed/mbed-os-example-tls/pull/63/files#diff-39d7102cc3f72bf4a2240cc439f38d0fR388
This also avoids uninitialised use.

tls-client/HelloHttpsClient.cpp
```diff
@@ -0,0 +1,431 @@
...
+    uint32_t flags;
+    bool resp_200, resp_hello;
```
mazimkhan: same as above.

tls-client/HelloHttpsClient.cpp
```diff
@@ -0,0 +1,431 @@
...
+    mbedtls_printf( "%s:%d: |%d| %s\r", basename, line, level, str );
+}
+
+int HelloHttpsClient::sslVerify( void *data, mbedtls_x509_crt *crt, int depth,
+                                 uint32_t *flags )
+{
+    (void)data;
+
+    int ret = -1;
+    char *buf = new char[GENERAL_PURPOSE_BUFFER_LENGTH];
```
mazimkhan: Is there a reason we cannot use ```gp_buf```` here? It is fine if we decide to use a localised ```gp_buf```` at the other place.

tls-client/HelloHttpsClient.cpp
```diff
@@ -0,0 +1,431 @@
...
+
+    return( ret );
+}
+
+void HelloHttpsClient::sslDebug( void *ctx, int level, const char *file,
+                                 int line, const char *str )
+{
+    (void)ctx;
+
+    const char *p, *basename;
```
mazimkhan: Uninitialised variables. If below loop fails to find ```basename``` then we could supplying a garbage value to the print functions below.

tls-client/HelloHttpsClient.cpp
```diff
@@ -0,0 +1,431 @@
...
+    /* Extract basename from file */
+    for( p = basename = file; *p != '\0'; p++ )
+    {
+        if( *p == '/' || *p == '\\' )
+        {
+            basename = p + 1;
+        }
+    }
+
+    mbedtls_printf( "%s:%d: |%d| %s\r", basename, line, level, str );
```
mazimkhan: Check if ```basename``` is valid by initializing it to NULL and checking here.

mazimkhan:
In general example code should be concise and quick to grasp. Scattering code should be avoided. Only take away for the users is the API usage.

andresag01:
Closing as this PR was created against the development branch, which will be deleted.

mazimkhan:
@andresag01 I think you were referring to this comment of mine https://github.com/ARMmbed/mbed-os-example-tls/pull/63#issuecomment-285700683
This comment is not against splitting a class but rather using a class for demonstrating use of a C library. The reason why I opposed it was that I believe an example should focus on the API that we want to demonstrate than adding extra things that can convolute the code and hence make it hard to follow by the intended users. In past there were some worst examples in BLE repo.
cc @Patater

Issue #69, opened 2017-03-16
wrong output in the Benchmark application


RonEld:
The output of the Benchmark Application performance is in Kb/s , however, according to the code:
~~~
    mbedtls_printf( "%9lu Kb/s,  %9lu cycles/byte\r\n",                        \
                     i * BUFSIZE / 1024,                                       \
                     ( mbedtls_timing_hardclock() - tsc ) / ( j * BUFSIZE ) );
~~~
The performance should be in Kilo Bytes (KB) not Kilo bits (Kb) as printed

ciarmcom:
ARM Internal Ref: IOTSSL-1286

andresag01:
A fix for this issue has already been merged into master. Closing ticket...

RonEld:
@andresag01 I think this issue was remained opened because of the benchmark application in the mbed TLS repository: https://github.com/ARMmbed/mbedtls/pull/920

RonEld:
my mistakes, I thought about the internal JIRA ticket

PR #71, opened 2017-03-16
Typo in Throughput log

RonEld:
Throughput is printed in Kb instead of KB
Fixed issue #69

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -13,21 +13,21 @@ You can also compile this example with the [mbed Online Compiler](https://develo
 The output in the terminal window should be similar to this:
 
 ```
-  SHA-256                  :       1673 Kb/s,         70 cycles/byte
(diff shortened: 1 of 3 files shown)


andresag01:
@RonEld: Thanks for spotting and fixing this issue. While the fix is good, I cannot merge this yet as it breaks the CI. The problem is that the print logs with `KB/s` no longer match the regular expression expected by `htrun` according to the test logs (refer to [this ](https://github.com/ARMmbed/mbed-os-example-tls/blob/master/tests/benchmark.log#L1)). Could you please update the test logs so that the CI passes?

andresag01:
@RonEld: Thank you for fixing the log regex log file. I'll wait for the CI to finish before merging.

andresag01:
All the expected test cases pass in the CI, merging this change.

Issue #74, opened 2017-03-29
[Benchmark] Remove dependencies on DWT registers


MarceloSalazar:
The application should not access to DWT/Debug registers directly. Also, these registers are only available in ARMv7-M based processors (not in Cortex-M0/M0+).

https://github.com/ARMmbed/mbed-os-example-tls/blob/master/benchmark/main.cpp#L189

I'd suggest to use the 'time' API available in mbed OS:
https://docs.mbed.com/docs/mbed-os-api-reference/en/latest/APIs/tasks/Time/

@sbutcher-arm

ciarmcom:
ARM Internal Ref: IOTSSL-1302

andresag01:
https://github.com/ARMmbed/mbed-os-example-tls/pull/88 addresses this issue.

andresag01:
PR #88 has now been merged. This issue should be resolved.

PR #76, opened 2017-03-29
benchmark: fix build error on Cortex-M0

gilles-peskine-arm:
The timing code uses the DWT cycle counter which is an M3/M4 feature.
For the sake of M0, use the mbed microsecond timer if DWT is not available.

This commit addresses #74

### DIFF
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -204,6 +204,7 @@ static unsigned long mbedtls_timing_before( void )
 {
     timer.reset( );
     timer.start( );
+    return( 0 );
 }
 static unsigned long mbedtls_timing_after( void )
 {
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -179,6 +179,8 @@
         mbedtls_printf( "FAILED: -0x%04x\r\n", -ret );
 #endif
 
+#if defined(CoreDebug_DEMCR_TRCENA_Msk) && defined(DWT_CTRL_CYCCNTENA_Msk)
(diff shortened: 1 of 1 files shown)


MarceloSalazar:
Please submit PR to mbed-os-workshop-17q2 branch (so we can use it this week).

benchmark/main.cpp
```diff
@@ -179,6 +179,8 @@
         mbedtls_printf( "FAILED: -0x%04x\r\n", -ret );
 #endif
 
+#if defined(CoreDebug_DEMCR_TRCENA_Msk) && defined(DWT_CTRL_CYCCNTENA_Msk)
+/* DWT cycle counter is available (Cortex-M3, Cortex-M4) */
```
MarceloSalazar: We shouldn't use different ways of measuring time. Using the same implementation will let us compare results across different platforms/MCUs.

gilles-peskine-arm: The code always prints kB/s. I only changed the additional printing of cycles/byte, which aren't meaningful across platforms. The only fully portable alternative is not to have a cycle count at all, AFAICT. So this is the best I know how to do at the moment, but I'm open to suggestions.

MarceloSalazar: I suggest we remove the usage of the cycle counter and use only one way of measuring time (which is supported across platforms).

gilles-peskine-arm: https://github.com/ARMmbed/mbed-os-example-tls/pull/79 removes the cycle counter altogether.

gilles-peskine-arm:
Rebased against workshop branch: PR #77

PR #77, opened 2017-03-29
Issue 74/use mbed timer 17q2 workshop

gilles-peskine-arm:
benchmark: fix build error on Cortex-M0

The timing code uses the DWT cycle counter which is an M3/M4 feature.
For the sake of M0, use the mbed microsecond timer if DWT is not available.
Throughput in kB/s remains shown on all platforms.

This commit addresses #74


andresag01:
Closing this PR as we have decided in favour of the fix at https://github.com/ARMmbed/mbed-os-example-tls/pull/79

PR #79, opened 2017-03-29
benchmark: remove cycle counter

gilles-peskine-arm:
The DWT cycle counter is only available on Cortex-M3 and Cortex-M4, so
the code didn't even build for Cortex-M0 targets. Now the code only
shows throughput in KB/s.

This commit addresses #74 

This PR is an alternative to https://github.com/ARMmbed/mbed-os-example-tls/pull/77 which retains the cycle counter where available.

### DIFF
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -179,25 +179,12 @@
         mbedtls_printf( "FAILED: -0x%04x\r\n", -ret );
 #endif
 
-static unsigned long mbedtls_timing_hardclock( void )
-{
-    static int dwt_started = 0;
-
-    if( dwt_started == 0 )
-    {
-        CoreDebug->DEMCR |= CoreDebug_DEMCR_TRCENA_Msk;
-        DWT->CTRL |= DWT_CTRL_CYCCNTENA_Msk;
-    }
-
-    return( DWT->CYCCNT );
-}
-
 static volatile int alarmed;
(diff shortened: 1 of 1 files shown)


MarceloSalazar:
LGTM

benchmark/main.cpp
```diff
@@ -207,15 +194,8 @@ do {                                                                           \
...
-    for( j = 0; j < 1024; j++ )                                                \
-    {                                                                          \
-        CODE;                                                                  \
-    }                                                                          \
-                                                                               \
-    mbedtls_printf( "%9lu KB/s,  %9lu cycles/byte\r\n",                        \
-                     i * BUFSIZE / 1024,                                       \
-                     ( mbedtls_timing_hardclock() - tsc ) / ( j * BUFSIZE ) ); \
+    mbedtls_printf( "%9lu KB/s\r\n",                                           \
+                     i * BUFSIZE / 1024 );                                     \
```
mazimkhan: Since we are changing prints we should also update examples log file for CI to pass https://github.com/ARMmbed/mbed-os-example-tls/blob/master/tests/benchmark.log

andresag01:
Closing this PR as https://github.com/ARMmbed/mbed-os-example-tls/issues/88 is a more complete version of the changes that includes docs and Ci logs updates.

PR #82, opened 2017-04-27
Updating mbed-os to mbed-os-5.4.4

0xc0170:
Please test/merge this PR and then tag Master with mbed-os-5.4.4


andresag01:
The change passes all the expected tests. Merging...

PR #87, opened 2017-06-07
Refactor tls-client example to improve readability

andresag01:
This patch completely refactors the tls-client example. The following
is a list of significant changes:
* Split the code into .h and .cpp files.
* Move the class HelloHTTPS to its own file.
* Remove all unused variables and redundant buffers.
* Split functionality into smaller, more manageable functions.
* Improve error reporting and logging.

### DIFF
--- a/tls-client/HelloHttpsClient.cpp
+++ b/tls-client/HelloHttpsClient.cpp
@@ -1,0 +1,431 @@
+/*
+ *  Hello world example of a TLS client: fetch an HTTPS page
+ *
+ *  Copyright (C) 2006-2017, ARM Limited, All Rights Reserved
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *  not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  This file is part of mbed TLS (https://tls.mbed.org)
+ */
+
+#include "HelloHttpsClient.h"
+
+#include "NetworkInterface.h"
+#include "TCPSocket.h"
+
+#include "mbedtls/platform.h"
+#include "mbedtls/config.h"
+#include "mbedtls/ssl.h"
+#include "mbedtls/entropy.h"
+#include "mbedtls/ctr_drbg.h"
+#include "mbedtls/error.h"
+#include "mbedtls/debug.h"
+
+#include <stdint.h>
+#include <string.h>
+
+HelloHttpsClient::HelloHttpsClient( const char *in_server_name,
+                                    const uint16_t in_server_port ) :
+    eth_iface( NULL ),
+    socket( NULL ),
+    server_name( in_server_name ),
+    server_port( in_server_port )
+{
+    mbedtls_entropy_init(&entropy);
+    mbedtls_ctr_drbg_init(&ctr_drbg);
+    mbedtls_x509_crt_init(&cacert);
+    mbedtls_ssl_init(&ssl);
+    mbedtls_ssl_config_init(&ssl_conf);
+}
+
+HelloHttpsClient::~HelloHttpsClient( void )
+{
+    mbedtls_entropy_free(&entropy);
+    mbedtls_ctr_drbg_free(&ctr_drbg);
+    mbedtls_x509_crt_free(&cacert);
+    mbedtls_ssl_free(&ssl);
+    mbedtls_ssl_config_free(&ssl_conf);
+
+    if( eth_iface != NULL )
+        delete eth_iface;
+    if( socket != NULL )
+        delete socket;
+}
+
+int HelloHttpsClient::createTCPSocket( void )
+{
+    int ret;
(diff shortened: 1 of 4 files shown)


andresag01:
This PR is equivalent to https://github.com/ARMmbed/mbed-os-example-tls/pull/63, but against the master branch.

andresag01:
I have reworked the PR according to the comments in #63 and changed the style to be closer to mbed OS.

tls-client/HelloHttpsClient.cpp
```diff
@@ -0,0 +1,381 @@
...
+    mbedtls_printf("Starting the TLS handshake...\r\n");
+    do {
+        ret = mbedtls_ssl_handshake(&ssl);
+    } while(ret != 0 &&
+            (ret == MBEDTLS_ERR_SSL_WANT_READ ||
+            ret == MBEDTLS_ERR_SSL_WANT_WRITE));
+    if (ret < 0) {
+        mbedtls_printf("mbedtls_ssl_handshake() returned -0x%04X\r\n", -ret);
+        return ret;
+    } else {
```
mazimkhan: No need for ```else``` if ```if``` block has a return. Otherwise it looks as following code is executed regardless. Applies at all places.

tls-client/HelloHttpsClient.cpp
```diff
@@ -0,0 +1,381 @@
...
+        resp_200 = resp_200 || strstr(gp_buf, HTTP_OK_STR) != NULL;
+        resp_hello = resp_hello || strstr(gp_buf, HTTP_HELLO_STR) != NULL;
+    } while((!resp_200 || !resp_hello) &&
+            (ret > 0 ||
+            ret == MBEDTLS_ERR_SSL_WANT_READ || MBEDTLS_ERR_SSL_WANT_WRITE));
+    if (ret < 0) {
+        mbedtls_printf("mbedtls_ssl_read() returned -0x%04X\r\n", -ret);
+        return ret;
+    }
+    gp_buf[resp_offset] = '\0';
```
mazimkhan: Already done at line https://github.com/ARMmbed/mbed-os-example-tls/pull/87/files#diff-39d7102cc3f72bf4a2240cc439f38d0fR208

andresag01:
@mazimkhan: Thanks for the review comments. I have changed the code accordingly.

mazimkhan:
retest CI

andresag01:
This PR needs to be updated to change the server name to `os.mbed.com`

tls-client/HelloHttpsClient.cpp
```diff
@@ -49,216 +84,302 @@ HelloHttpsClient::HelloHttpsClient( const char *in_server_name,
...
 
-    ip_addr = eth_iface->get_ip_address();
-    if( ip_addr != NULL )
-    {
-        mbedtls_printf( "Client IP address is %s\r\n", ip_addr );
+    /* Fill the request buffer */
+    ret = snprintf(gp_buf, GENERAL_PURPOSE_BUFFER_LENGTH - 1,
+                   "GET %s HTTP/1.1\nHost: %s\n\n", HTTP_REQUEST_FILE_PATH,
+                   server_name);
+    req_len = static_cast<size_t>(ret);
```
k-stachowiak: This cast is redundant - it will be performed during the assignment

tls-client/main.cpp
```diff
@@ -19,417 +19,72 @@
...
-
-#if DEBUG_LEVEL > 0
-    /**
-     * Debug callback for mbed TLS
-     * Just prints on the USB serial port
-     */
-    static void my_debug(void *ctx, int level, const char *file, int line,
-                         const char *str)
+    /* Allocate a HTTPS client */
+    client = new HelloHttpsClient( SERVER_NAME, SERVER_PORT );
```
k-stachowiak: Unless there is a reason (e.g. stack memory limit) to allocate the client in the free sore (the heap) I'd advise stack allocation. This will enable using plain return statements below instead of storing the exit code in a variable, which you need to track when reading the code.

andresag01: @k-stachowiak: In the past this structure was allocated in the stack, but as you said, we ran into a lot of stack overflow problems. For this reason it is allocated in the heap...

tls-client/HelloHttpsClient.cpp
andresag01: It is cast implicitly, but I added the explicit cast to ensure that no signed to unsigned implicit cast errors are generated by the compiler...

andresag01:
The last few commits update this PR to integrate the new version of tls-client with easy-connect, update the README.md sample output and the regular expressions for htrun in the CI. I have also triggered another CI run.

andresag01:
@Patater: This PR passes all the expected tls-client tests. Could you please take a look? The majority of changes have already been reviewed. The only thing that was not reviewed is the integration with easy-connect, but those changes are small. Perhaps we just need gatekeeper review?

Patater:
@andresag01 Could you please rebase and resolve conflicts?

andresag01:
@Patater: I have rebased the changes. Waiting for the CI to run to see if there are any regressions.

andresag01:
@Patater: I checked the test results and the only failure is `benchmark-KW24D-IAR` apparently due to an allocation failure. Given that this change is unrelated to the benchmark example, I think this PR is ok.

andresag01:
@Patater: I have checked the test results and there seem to be some unrelated failures in the hashing and benchmark applications. Also, it `build-tls-client-UBLOX_EVK_ODIN_W2-GCC_ARM` failed with the following error:
```
socket.connect() returned -3007
```
But I am not sure what the issue is because it works with GCC and I do not have the board to reproduce the problem (if any). Also, note that the same test case worked in this [previous CI run](https://jenkins-internal.mbed.com/job/mbedtls-mbed-os-examples-pr/94/flowGraphTable/) and I have not made any code changes since then...

Issue #91, opened 2017-06-08
Authcrypt compiled with make_gcc_arm fails with runtime error


bridadan:
With branch `mbed-os-5.5.0-rc1-oob`, when exporting with `make_gcc_arm`, the example fails to run.

I received two different errors when running the authcrypt example.

```
plaintext message: 536f6d65207468696e67732061726520626574746572206c65667420756e7265616400
CMSIS-RTOS error: Stack underflow (status: 0x1, task ID: 0x20006CE8, task name: (null))
```

```
plaintext message: 536f6d65207468696e67732061726520626574746572206c65667420756e7265616400
Mutex 0x730fcfdb error -4: Parameter error
```

deepikabhavnani:
@bridadan on which target device?

yennster:
I received a similar error on HEXIWEAR with the following commands: `mbed export -i make_gcc_arm -m HEXIWEAR` & `make` -

```
plaintext message: 536f6d65207468696e67732061726520626574746572206c65667420756e7265616400
CMSIS-RTOS error: Stack underflow (status: 0x1, task ID: 0x20003C00, task name: (null))
```

yennster:
@deepikabhavnani @bridadan Once again, I received a similar error on K66F with the following commands: `mbed export -i make_gcc_arm -m K66F` & `make` -

```
plaintext message: 536f6d65207468696e67732061726520626574746572206c65667420756e7265616400
CMSIS-RTOS error: Stack underflow (status: 0x1, task ID: 0x20003400, task name: (null))
```

andresag01:
I tried with both HEXIWEAR and K64F and the example fails in both cases when the application is compiled with the debug profile. For example, compiling with the following command will result in the same stack underflow error:
```
mbed compile -t GCC_ARM -m K64F --profile mbed-os/tools/profiles/debug.json
```

bridadan:
Sorry y'all, it was late on a Thursday, wasn't thinking straight. It was on the UBLOX_EVK_ODIN_W2.

@andresag01 good catch about the debug profile, maybe this is a generic issue and not just for the TLS examples...

andresag01:
It seems that we are running out of stack memory and a runtime check seems to be throwing this failure. As I pointed out before, the problem only seems to happen with debug builds, while no error occurs with release or develop builds probably because in debug builds compiler optimisations are completely disabled.

[...]

PR #95, opened 2017-06-13
Refactor authcrypt example

andresag01:
The example is refactored to use C++ classes and allocate the bulk of
the mbed TLS context structures in the heap instead of stack.

This change should fix https://github.com/ARMmbed/mbed-os-example-tls/issues/91.

### DIFF
--- a/authcrypt/authcrypt.cpp
+++ b/authcrypt/authcrypt.cpp
@@ -1,0 +1,185 @@
+/*
+ *  Hello world example of using the authenticated encryption with mbed TLS
+ *
+ *  Copyright (C) 2017, ARM Limited, All Rights Reserved
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *  not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include "authcrypt.h"
+
+#include "mbed.h"
+
+#include "mbedtls/cipher.h"
+#include "mbedtls/entropy.h"
+#include "mbedtls/ctr_drbg.h"
+#if DEBUG_LEVEL > 0
+#include "mbedtls/debug.h"
+#endif
+
+#include "mbedtls/platform.h"
+
+#include <string.h>
+
+const unsigned char Authcrypt::secret_key[16] = {
+    0xf4, 0x82, 0xc6, 0x70, 0x3c, 0xc7, 0x61, 0x0a,
+    0xb9, 0xa0, 0xb8, 0xe9, 0x87, 0xb8, 0xc1, 0x72,
+};
+
+const char Authcrypt::message[] = "Some things are better left unread";
+
+const char Authcrypt::metadata[] = "eg sequence number, routing info";
+
+Authcrypt::Authcrypt()
+{
+    memset(ciphertext, 0, sizeof(ciphertext));
+    memset(decrypted, 0, sizeof(decrypted));
+
+    mbedtls_entropy_init(&entropy);
+    mbedtls_ctr_drbg_init(&drbg);
+    mbedtls_cipher_init(&cipher);
+}
+
+Authcrypt::~Authcrypt()
+{
+    memset(ciphertext, 0, sizeof(ciphertext));
+    memset(decrypted, 0, sizeof(decrypted));
+
+    mbedtls_cipher_free(&cipher);
+    mbedtls_ctr_drbg_free(&drbg);
+    mbedtls_entropy_free(&entropy);
+}
+
+int Authcrypt::run()
+{
+    mbedtls_printf("\r\n\r\n");
+    print_hex("plaintext message",
+              reinterpret_cast<const unsigned char *>(message),
+              sizeof(message));
+
+    /*
+     * Seed the PRNG using the entropy pool, and throw in our secret key as an
+     * additional source of randomness.
+     */
+    int ret = mbedtls_ctr_drbg_seed(&drbg, mbedtls_entropy_func, &entropy,
+                                    secret_key, sizeof(secret_key));
+    if (ret != 0) {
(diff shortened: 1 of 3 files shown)


authcrypt/main.cpp
```diff
@@ -53,132 +39,174 @@ static const unsigned char secret_key[16] = {
...
+    /* Pseudo-random generator */
+    mbedtls_ctr_drbg_context drbg;
+    mbedtls_cipher_context_t cipher;
+};
 
 int main() {
-    int ret = example();
-    if (ret != 0) {
+    int ret;
+    Authcrypt *authcrypt = new Authcrypt();
```
RonEld: check if the heap is contigous. If not, the `ciphertext` and `decrypted` will not be properly read by HW acceleration engines

andresag01:
@RonEld: I have fixed some of the issues you pointed out. However, I do not think that we should split the source file into .h and .c as this is a fairly simple class.

RonEld:
@andresag01 thanks for addressing my comments
I think that although it is a simple class, it is not so readable, now that the implementation and definition are together. 
In addition, without the split, there is no data abstraction and encapsulation here

authcrypt/main.cpp
```diff
@@ -53,132 +39,177 @@ static const unsigned char secret_key[16] = {
...
+                        reinterpret_cast<const unsigned char *>(metadata),
+                        sizeof(metadata), ciphertext + nonce_len,
+                        ciphertext_len - nonce_len - tag_len, decrypted,
+                        &decrypted_len, ciphertext + ciphertext_len - tag_len,
+                        tag_len);
+        /* Checking the return code is CRITICAL for security here */
+        if (ret == MBEDTLS_ERR_CIPHER_AUTH_FAILED) {
+            mbedtls_printf("Something bad is happening! Data is not "
+                           "authentic!\r\n");
+            return 1;
```
mazimkhan: Why are we changing return from ```ret``` to ```0```? At https://github.com/ARMmbed/mbed-os-example-tls/pull/95/files#diff-64da713f1246019fb29c1e604285e6e0R211 the return code is printed again and will always have ```1```.
Applies to all places returning 1.

authcrypt/main.cpp
```diff
@@ -53,132 +39,177 @@ static const unsigned char secret_key[16] = {
...
-    /* Seed the PRNG using the entropy pool, and throw in our secret key as an
-     * additional source of randomness. */
-    ret = mbedtls_ctr_drbg_seed(&drbg, mbedtls_entropy_func, &entropy,
-                                       secret_key, sizeof secret_key);
-    if (ret != 0) {
-        return 1;
+/* Message that should be protected */
+static const char message[] = "Some things are better left unread";
+/* Metadata transmitted in the clear but authenticated */
+static const char metadata[] = "eg sequence number, routing info";
```
mazimkhan: Any particular reason for leaving ```message```, ```metadata``` and ```secret_key``` outside the class. If this data belongs to this class then it should be encapsulated inside this class.

authcrypt/main.cpp
```diff
@@ -53,132 +39,177 @@ static const unsigned char secret_key[16] = {
...
+    /* Pseudo-random generator */
+    mbedtls_ctr_drbg_context drbg;
+    mbedtls_cipher_context_t cipher;
+};
 
 int main() {
-    int ret = example();
-    if (ret != 0) {
+    int ret;
+    Authcrypt *authcrypt = new Authcrypt();
```
mazimkhan: ```new``` can fail to allocate the memory. We can assert or simply print error if we can run the example.

mazimkhan:
On splitting the code I agree with Ron. When we look at C++ code we want to see two things
- How the class is defined i.e. its data, methods, permissions etc.
- Method implementation
With everything together it is hard to view the class itself.

authcrypt/main.cpp
andresag01: Ahh thanks for spotting that. This is because I did not modify the original code, just changed the memory allocation. I will change this as well.

andresag01:
@mazimkhan, @RonEld: I have modified the example according to your comments. Could you please take a look again?

mazimkhan:
Can you please also put a NULL check here https://github.com/ARMmbed/mbed-os-example-tls/pull/95/commits/b3cebca6c23978a3cb09715f7c8df410a261fe0d#diff-64da713f1246019fb29c1e604285e6e0R28

authcrypt/authcrypt.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include <string.h>
+
+const unsigned char Authcrypt::secret_key[16] = {
+    0xf4, 0x82, 0xc6, 0x70, 0x3c, 0xc7, 0x61, 0x0a,
+    0xb9, 0xa0, 0xb8, 0xe9, 0x87, 0xb8, 0xc1, 0x72,
+};
+
+const char Authcrypt::message[] = "Some things are better left unread";
+
+const char Authcrypt::metadata[] = "eg sequence number, routing info";
```
RonEld: why not initialize message and metadata in the CTOR:
~~~
Authcrypt():message("Some things are better left unread"),metadata("eg sequence number, routing info")
~~~
This way, these members would remain private

andresag01:
@mazimkhan: I looked into this at some point, but I do not think new returns null: http://www.cplusplus.com/reference/new/operator%20new/. I believe the normal behaviour is to throw an exception if allocation is not possible. However, exceptions are disabled in mbed OS. An alternative could be to use nothrow, but that is not enabled in mbed OS either...

mazimkhan:
Referring to my previous comment https://github.com/ARMmbed/mbed-os-example-tls/pull/95#issuecomment-308727256

Just check whether it throws bad alloc if new fails or returns NULL. For former we don't need to do anything.

authcrypt/authcrypt.cpp
andresag01: These members are already private. They are under the `private` section of the class declaration.

RonEld: message and metadata are static const members, in the public section

andresag01:
@mazimkhan: I do not think there is anything we can do. If new fails we just die, check here: https://github.com/ARMmbed/mbed-os/blob/master/platform/mbed_retarget.cpp#L941

authcrypt/authcrypt.cpp
andresag01: Thanks for pointing it out that they were public, I have now made them private members.

mazimkhan:
Thanks for looking into this.

mazimkhan:
LGTM

andresag01:
Rebase changes to run the CI with the latest mbed OS version 5.5.2

andresag01:
The change has been approved and the relevant tests pass in the CI, merging this PR.

PR #109, opened 2017-08-02
tls-client: Check certificate verification flags to exclude time failures

andresag01:
This patch is necessary when the
MBEDTLS_HAVE_TIME_DATE functionality is enabled because the Real-Time
Clock needs to be manually set to the correct time, which is not a
trivial task in this example code. The idea here is to configure the certificate verification function to filter out date/time failures during the TLS handshake.

**NOTES:**
* @sbutcher-arm: This PR fixes the tls-client example when [this mbed OS PR](https://github.com/ARMmbed/mbed-os/pull/4846) is merged.

### DIFF
--- a/tls-client/HelloHttpsClient.cpp
+++ b/tls-client/HelloHttpsClient.cpp
@@ -30,6 +30,7 @@
 #include "mbedtls/ctr_drbg.h"
 #include "mbedtls/error.h"
 #include "mbedtls/debug.h"
+#include "mbedtls/x509.h"
 
 #include <stdint.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


sbutcher-arm:
@k-stachowiak - Can you please test this PR, review it and approve it if possible?

tls-client/HelloHttpsClient.cpp
```diff
@@ -363,9 +366,18 @@ void HelloHttpsClient::sslDebug(void *ctx, int level, const char *file,
...
+    /*
+     * If MBEDTLS_HAVE_TIME_DATE is defined, then the certificate date and time
+     * validity checks will probably fail because this application does not set
+     * up the clock correctly. We filter out date and time related failures
+     * instead
+     */
+    *flags &= ~MBEDTLS_X509_BADCERT_FUTURE & ~MBEDTLS_X509_BADCERT_EXPIRED;
 
-    int ret = -1;
+#if HELLO_HTTPS_CLIENT_DEBUG_LEVEL > 0
```
sbutcher-arm: @mazimkhan / @andresag01 - Will changing what's displayed as debug impact the CI, which uses text output to validate the example?

andresag01: This does not change what is displayed on the terminal unless the HELLO_HTTPS_CLIENT_DEBUG_LEVEL macro is > 0 (which I think isnt by default). In fact, the  changes in the placing of the macro is to ensure this happens. Note that in the past the sslVerify function cas not configured unless HELLO_HTTPS_CLIENT_DEBUG_LEVEL> 0. However, now it is because we need it to clear the timing flags. Hence, I had to put extra macro checks around the printing code in sslVerify() to prevent it from printing by default.

cc @k-stachowiak: I think this answers your question as well.

mazimkhan: If this change does not modify the prints that appear in `mbed-os-example-tls/tests/tls-client.log` then this will not impact the CI. Any additional prints are ignored.

mazimkhan:
retest

mazimkhan:
retest

mazimkhan:
retest

mazimkhan:
CI fails due to a known issue #138. Logs: https://jenkins-internal.mbed.com/view/mbed-tls/job/mbedtls-mbed-os-examples-pr/138/execution/node/632/log/

PR #112, opened 2017-08-17
Disable AES hardware acceleration for STM32F439xI

andresag01:
@Patater @yanesca


andresag01:
This PR is a work around for the following mbed OS issue: https://github.com/ARMmbed/mbed-os/issues/4928

andresag01:
All the expected tests pass. Merging these changes...

JanneKiiskila:
Would it have been easier to modify the targets from mbed OS itself?

PR #125, opened 2017-11-02
Change Arm trademarks in README.md files

andresag01:
Please note that this PR only changes the branding in the README.md files like in https://github.com/ARMmbed/mbedtls/pull/1159 but does not change any of the branding in the actual source code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# mbed TLS Examples on mbedOS
+# Mbed TLS Examples on Mbed OS
 
-This repository contains a collection of mbed TLS example applications based on mbed OS. Each subdirectory contains a separate example meant for building as an executable.
+This repository contains a collection of Mbed TLS example applications based on Mbed OS. Each subdirectory contains a separate example meant for building as an executable.
 
 # Getting started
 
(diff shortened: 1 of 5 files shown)


andresag01:
@RonEld @Patater @sbutcher-arm: I noticed that there are a few occurrences of the previous branding in this repo. Could some of you check that I did the correct thing before merging? Thanks!

Patater:
If the approach we want to take is to only update README files, your changes LGTM.

However, there isn't too much source code comments with "mbed" or "ARM", compared to the main mbedtls repo. It'd be a good idea to update those as well, (and I'm not just thinking about copyright strings, but use of the brand in comments like "Hello world example of using the hashing functions of mbed TLS"). Examples are often one of the first interactions with brand, so it makes sense to get it right in examples.

sbutcher-arm:
This PR LGTM.

andresag01:
I have marked this as "ready for merge" as the PR has been approved by three different reviewers. Given that this PR only contains documentation changes in README.md files, I do not think it is necessary to run the full CI.

Patater:
Reviewed, approved and merged.

PR #136, opened 2017-12-13
Dummy PR to test in CI

mazimkhan:


Patater:
Removing as we no longer need this PR to test the CI

PR #146, opened 2018-02-04
Add heap statistics to tls-example

JanneKiiskila:
This highlights the heap memory usage in an easy-to-use way.
You can quite clearly see how the heap usage peaks up.

### DIFF
--- a/tls-client/README.md
+++ b/tls-client/README.md
@@ -92,6 +92,33 @@ Another possible reason for this error is a proxy providing a different certific
 
 **Warning:** this removes all security against a possible active attacker, so use at your own risk or for debugging only!
 
+## Heap statistics
+
+If you compile the application with `DHEAP_STATS_ENABLED=1` option will enable the heap statistics printouts. 
+
+Mbed OS feature a heap statistics monitoring feature and you can measure the heap memory consumption (current and maximum). You can then clearly see from the printouts how much heap memory is consumed by the operations.
+
+Example printout looks like this:
+
+```
+Starting mbed-os-example-tls/tls-client
+**** heap_stats at Starting
+**** current_size: 1024
+**** max_size : 1024
+```
+
+at startup we consume 1 kilobyte of heap. 
+
+```
+**** heap_stats at Finished
+**** current_size: 2960
+**** max_size : 53131
+```
+
+However, at the end we can see our peak heap consumption was over 53 kilobytes.
+
+More information on Mbed OS memory statistics available in [documentation](https://docs.mbed.com/docs/mbed-os-handbook/en/5.2/advanced/runtime_stats/).
+
 ## Troubleshooting
 
 If you have problems, you can review the [documentation](https://os.mbed.com/docs/latest/tutorials/debugging.html) for suggestions on what could be wrong and how to fix it.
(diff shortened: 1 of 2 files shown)


k-stachowiak:
@JanneKiiskila I find your suggestion of using the Mbed OS heap usage measurement technique interesting. What do you think about introducing it in the benchmark example application? It seems to me like a more appropriate place than the tls-client. To me there is no special connection between the tls-client application and the memory usage measurement, whereas the benchmark example has been created exactly for that purpose. Also note that it uses the Mbed TLS measurement API rather than the Mbed OS one, so incorporating your suggested method into the benchmark example could prove more instructive there than in this application. It could be selectable in the runtime to demonstrate both approaches and enable their comparison.

andresag01:
@JanneKiiskila: Thanks a lot for the contribution. While I agree that heap memory profiling is important, I think it would be best to try to keep the example Mbed TLS applications limited to demonstrating Mbed TLS features, but I am definitely in favour of adding a section in the README pointing to the Mbed OS documentation for heap and stack stats. What do you think?

@k-stachowiak: I see what you mean regarding the benchmark application, but as I mentioned above, I think we should limit the Mbed TLS examples to demonstrating the Mbed TLS features and avoid increasing complexity. Also, note that the benchmark application is using very different features from the TLS client, and it seems that @JanneKiiskila  is mostly interested in the latter, so adding the stats to benchmark might not help much.

JanneKiiskila:
Well, I'm not fully agreeing here. The Benchmark does not do any kind of TLS/DTLS connection, so from my point of view (Mbed Cloud Client glasses on) - this benchmark does almost nothing that would  interest me. It might make sense to add it also there, but that is another PR.

From Client's point of view the MbedTLS peak RAM usage is the key issue.

JanneKiiskila:
If you think this complicates things too much (though it's relatively simple, but I'm tempted to add also dynamic stack usage tracing as well) -> it does not have to get merged in, it can reside also in it's own branch or in my fork, so that if anyone's interested - they can use it.

JanneKiiskila:
Complexity +1 new function (very trivial), 
+6 function calls to main - of which we could drop some of the intermediate ones and just leave the 1st and last to make it a tad bit simpler?

andresag01:
@JanneKiiskila: As I mentioned in my earlier comment, I agree that adding the heapstats or stackstats into the benchmark example does not help much.

With regards to your other comment regarding the complexity. I am not too worried about this as it is only a few well placed lines of code. However, please bear in mind that this is an Mbed TLS example application to demonstrate the use of the TLS stack. If we add more and more features (simple as they might be) the example code will become more and more cluttered and it will probably become harder for newcomers to follow. In the interest of keeping the applications as simple and understandable as we can, then I personally prefer to not include functionality such as heap and stack stats.

Finally, as you mentioned, the TLS peak heap usage is very interesting for some people while developing with Mbed TLS and Mbed OS. Therefore, I suggest adding documentation in the example's README that points to the Mbed OS documentation which already contains code to do most of what is in the PR. For example: https://docs.mbed.com/docs/mbed-os-handbook/en/5.2/advanced/runtime_stats/.

What do you think?

Patater:
I am closing this PR in order to keep the examples razor-focused on Mbed TLS features and not on Mbed OS features like heap stats. While this proposed change is simple and useful, the example really isn't the place for this sort of thing.

PR #150, opened 2018-02-28
Warn about improper error reporting by ATCAECC508A driver

mazimkhan:
ATCAECC508A driver may not report device initialisation errors properly when device not connected to the Mbed platform or it is not commissioned.
Warning added here is temporary and will be removed once driver is fixed. Driver is not fixed in the demo release now since the time to merge in mbed-os is longer.


mazimkhan:
Merging as trivial documentation changes.

PR #154, opened 2018-03-09
Add information about opaque keys examples

mazimkhan:
Added information about examples that demonstrate opaque keys. 
Moved examples overview from the monitoring section to the top.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,16 @@
 # Mbed TLS Examples on Mbed OS
 
-This repository contains following Mbed TLS example applications based on Mbed OS. Each subdirectory contains a separate example meant for building as an executable. Please browse to the subdirectories for specific documentation.
+This repository contains a collection Mbed TLS example applications based on Mbed OS. Each subdirectory contains a separate example meant for building as an executable. Please browse to the subdirectories for specific documentation.
 
 * [authcrypt](./authcrypt/README.md): performs authenticated encryption and authenticated decryption of a buffer.
 * [benchmark](./benchmark/README.md): benchmarks the various cryptographic primitives offered by Mbed TLS.
 * [hashing](./hashing/README.md): performs hashing of a buffer with SHA-256 using various APIs.
 * [tls-client](./tls-client/README.md): downloads a file from an HTTPS server (os.mbed.com) and looks for a specific string in that file.
 
-In addition following examples demonstrate the Mbed TLS opaque keys feature using hardware crypto engine [ATECC508A](https://github.com/ARMmbed/mbed-os/tree/feature-opaque-keys/features/atcryptoauth/README.md).
-* [pk_sign](./pk_sign/README.md): create signature using opaque keys and verify in software.
-* [pk_verify](./pk_verify/README.md): create and verify signature using opaque keys.
-* [tls-client-with-client-authentication](./tls-client-with-client-authentication/README.md): downloads a file from an HTTPS server with client authentication. The client authentication key can be an opaque key.
+In addition following examples demonstrate the use of Mbed TLS using a Secure IC and off-chip cryptographic accelerator [ATECC508A](https://github.com/ARMmbed/mbed-os/tree/feature-opaque-keys/features/atcryptoauth/README.md).
+* [pk_sign](./pk_sign/README.md): create signature using external Security IC and verify in software.
+* [pk_verify](./pk_verify/README.md): create and verify signature using external Security IC.
+* [tls-client-with-client-authentication](./tls-client-with-client-authentication/README.md): downloads a file from an HTTPS server with client authentication, using keys secured in an external Security IC.
 
 # Getting started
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,6 +36,8 @@ An alternative to Mbed CLI is to use the [Mbed Online Compiler](https://os.mbed.
 * [hashing](https://os.mbed.com/teams/mbed-os-examples/code/mbed-os-example-tls-hashing)
 * [tls-client](https://os.mbed.com/teams/mbed-os-examples/code/mbed-os-example-tls-tls-client)
 
+Note: examples pk_sign, pk_verify and tls-client-with-client-authentication are not available on the online IDE.
```
Patater: Why not? Is there an issue tracking this deficiency we could link to here?

mazimkhan: These examples can be made available as projects in online IDE. They are just not there.

sbutcher-arm: Change it to 'not yet available' then.

README.md
```diff
@@ -1,6 +1,16 @@
...
+
+* [authcrypt](./authcrypt/README.md): performs authenticated encryption and authenticated decryption of a buffer.
+* [benchmark](./benchmark/README.md): benchmarks the various cryptographic primitives offered by Mbed TLS.
+* [hashing](./hashing/README.md): performs hashing of a buffer with SHA-256 using various APIs.
+* [tls-client](./tls-client/README.md): downloads a file from an HTTPS server (os.mbed.com) and looks for a specific string in that file.
+
+In addition following examples demonstrate the Mbed TLS opaque keys feature using hardware crypto engine [ATECC508A](https://github.com/ARMmbed/mbed-os/tree/feature-opaque-keys/features/atcryptoauth/README.md).
+* [pk_sign](./pk_sign/README.md): create signature using opaque keys and verify in software.
+* [pk_verify](./pk_verify/README.md): create and verify signature using opaque keys.
+* [tls-client-with-client-authentication](./tls-client-with-client-authentication/README.md): similar to the tls-client example. Except it demonstrates client authentication and opaque keys.
```
gilles-peskine-arm: Grammar: you can't put a full stop before “except”. But it would be better to make the example stand on its own. Also this client does not work against `os.mbed.com` (because that wouldn't use client authentication). Suggestion:

> downloads a file from an HTTPS server with client authentication. The client authentication key can be an opaque key.

README.md
mazimkhan: Created issue https://github.com/ARMmbed/mbed-os-example-tls/issues/155. Will add it in the docs.

README.md
```diff
@@ -1,6 +1,16 @@
 # Mbed TLS Examples on Mbed OS
 
-This repository contains a collection of Mbed TLS example applications based on Mbed OS. Each subdirectory contains a separate example meant for building as an executable.
+This repository contains following Mbed TLS example applications based on Mbed OS. Each subdirectory contains a separate example meant for building as an executable. Please browse to the subdirectories for specific documentation.
```
sbutcher-arm: 'This repository contains following Mbed TLS example applications' doesn't make sense.
 
I think you need to revert to the original line:
> This repository contains a collection of Mbed TLS example applications based on Mbed OS.

README.md
```diff
@@ -1,6 +1,16 @@
...
+
+* [authcrypt](./authcrypt/README.md): performs authenticated encryption and authenticated decryption of a buffer.
+* [benchmark](./benchmark/README.md): benchmarks the various cryptographic primitives offered by Mbed TLS.
+* [hashing](./hashing/README.md): performs hashing of a buffer with SHA-256 using various APIs.
+* [tls-client](./tls-client/README.md): downloads a file from an HTTPS server (os.mbed.com) and looks for a specific string in that file.
+
+In addition following examples demonstrate the Mbed TLS opaque keys feature using hardware crypto engine [ATECC508A](https://github.com/ARMmbed/mbed-os/tree/feature-opaque-keys/features/atcryptoauth/README.md).
```
sbutcher-arm: I don't think we want to mention opaque keys. It's a meaningless term outside the team. May I suggest:

> In addition the following examples demonstrate the use of Mbed TLS using a Secure IC and off-chip cryptographic accelerator.

README.md
sbutcher-arm: 'similar to the tls-client example. Except it demonstrates client authentication and opaque keys.' needs to be one sentence, and again I think you should remove reference to opaque keys.

Can I suggest:
> similar to the tls-client example, this example demonstrates client authentication using keys secured in an external Security IC

mazimkhan:
@gilles-peskine-arm @Patater incorporated your suggestions.

README.md
mazimkhan: @sbutcher-arm addressed your comments.

mazimkhan:
@sbutcher-arm addressed your comments.

sbutcher-arm:
@Patater - Can we close this PR now, as the feature won't ship in this form?

mazimkhan:
Closing as feature requirements changed.

Patater:
The opaque keys feature will ship as part of [Mbed Crypto](https://github.com/ARMmbed/mbed-crypto)

PR #165, opened 2018-04-12
Add platform setup and teardown support

AndrzejKurek:
Add platform setup and teardown for authcrypt, benchmark, hashing and tls-client examples.
Tested on K64F with GCC_ARM.

### DIFF
--- a/authcrypt/main.cpp
+++ b/authcrypt/main.cpp
@@ -24,6 +24,11 @@
 #include "mbedtls/platform.h"
 
 int main() {
+    mbedtls_platform_context platform_ctx;
+    if(mbedtls_platform_setup(&platform_ctx)) {
+        return -1;
+    }
+
     int exit_code = MBEDTLS_EXIT_SUCCESS;
     Authcrypt *authcrypt = new Authcrypt();
 
@@ -34,5 +39,6 @@ int main() {
 
     delete authcrypt;
 
+    mbedtls_platform_teardown(&platform_ctx);
     return exit_code;
 }
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -961,8 +961,14 @@ static int benchmark( int argc, char *argv[] )
 }
 
 int main(void) {
+    mbedtls_platform_context platform_ctx;
+    if(mbedtls_platform_setup(&platform_ctx)) {
+        return -1;
+    }
+
     int ret = benchmark(0, NULL);
     if (ret != 0) {
         mbedtls_printf("Benchmark failed with error %d\r\n", ret);
     }
+    mbedtls_platform_teardown(&platform_ctx);
 }
--- a/hashing/main.cpp
+++ b/hashing/main.cpp
@@ -152,8 +152,14 @@ static int example(void)
 }
 
 int main() {
+    mbedtls_platform_context platform_ctx;
+    if( mbedtls_platform_setup(&platform_ctx)) {
+        return -1;
+    }
+
     int ret = example();
     if (ret != 0) {
         mbedtls_printf("Example failed with error %d\r\n", ret);
     }
+    mbedtls_platform_teardown(&platform_ctx);
 }
--- a/tls-client/main.cpp
+++ b/tls-client/main.cpp
@@ -412,6 +412,10 @@ protected:
  * The main loop of the HTTPS Hello World test
  */
 int main() {
+    mbedtls_platform_context platform_ctx;
+    if( mbedtls_platform_setup(&platform_ctx)) {
+        return -1;
+    }
     /* The default 9600 bps is too slow to print full TLS debug info and could
      * cause the other party to time out. */
 
@@ -432,10 +436,13 @@ int main() {
 #endif /* DEBUG_LEVEL > 0 */
     if (NULL == network) {
         printf("Connecting to the network failed... See serial output.\n");
+        mbedtls_platform_teardown(&platform_ctx);
         return 1;
     }
 
     HelloHTTPS *hello = new HelloHTTPS(HTTPS_SERVER_NAME, HTTPS_SERVER_PORT, network);
     hello->startTest(HTTPS_PATH);
     delete hello;
+    
+    mbedtls_platform_teardown(&platform_ctx);
 }


RonEld:
ALthough logically it is correct, I think it is better to have the calls where all the cryptography functions are, so `platform_ctx` can be accessible ( after https://github.com/ARMmbed/mbedtls/issues/1200 is resolved )

AndrzejKurek:
@RonEld Currently the platform context is not needed or used by the tests. Even if it were, it would be an external dependency for a test case (not a part of ones setup), so I'd opt for a dependency injection using a function parameter.

AndrzejKurek:
The CI fails on two known issues - memory allocation failure on one board, and lack of serial on another.

RonEld:
@AndrzejKurek I agree, but if you don't have access to `platform_ctx` in your test code(without having it external, which loses the idea of having it as a parameter ), how can you inject it as a function parameter?

AndrzejKurek:
@RonEld I don't really understand. If the platform context is somehow needed in the, for example, benchmark test, it can be passed to the function handling the test (and therefore where the cryptographic calls are made, as you suggested), instead of making it a global structure.

RonEld:
@AndrzejKurek I believe we are talking about the same thing.
I am asking the context to be defined in `benchmark()` function (in the benchmark example), and not in `main()`, as all the cryptographic functions are called within benchmark, and not within main.
I am not asking it to be global.

AndrzejKurek:
@RonEld And I'm saying that platform initialization is not part of the benchmark test, it is an "external" (from the benchmark test point of view) dependency, which could be satisfied using dependency injection - which would be creating it within main. This way, we can for example swap the platform context for a debug platform context, or extend it in any way we want to. This is a common pattern in, for example, unit testing, where objects or dependencies can be swapped with their respective mocks to perform the testing efficiently and at the relevant layer of abstraction.

RonEld:
OK, so let's compromise on having the test constructors getting the context as parameter? (or adding a "set platform context" function?). Without the tests themselves getting access to the context, there is no much use in having this context

authcrypt/main.cpp
```diff
@@ -24,6 +24,11 @@
 #include "mbedtls/platform.h"
 
 int main() {
+    mbedtls_platform_context platform_ctx;
+    if(mbedtls_platform_setup(&platform_ctx)) {
```
RonEld: I prefer explicitly checking result for `!=0`

benchmark/main.cpp
```diff
@@ -961,8 +961,14 @@ static int benchmark( int argc, char *argv[] )
 }
 
 int main(void) {
+    mbedtls_platform_context platform_ctx;
+    if(mbedtls_platform_setup(&platform_ctx)) {
```
RonEld: I prefer explicitly checking result for `!=0`

tls-client/main.cpp
```diff
@@ -412,6 +412,10 @@ class HelloHTTPS {
  * The main loop of the HTTPS Hello World test
  */
 int main() {
+    mbedtls_platform_context platform_ctx;
+    if( mbedtls_platform_setup(&platform_ctx)) {
```
RonEld: I prefer explicitly checking result for `!=0`

hashing/main.cpp
```diff
@@ -152,8 +152,14 @@ static int example(void)
 }
 
 int main() {
+    mbedtls_platform_context platform_ctx;
+    if( mbedtls_platform_setup(&platform_ctx)) {
```
RonEld: I prefer explicitly checking result for `!=0`

AndrzejKurek:
@RonEld The tests getting the context is the next step, but I'm hesitant to make this change without the tests doing anything with the context at all. Do You think that adding it, adding a (void) call in the beginning and commenting that it can be used if necessary will be OK?

RonEld:
I believe you will probably get compilation warnings for unused parameter, but we can handle this.
I prefer adding this call, so it would show intent, with adding a comment as you mentioned, referring https://github.com/ARMmbed/mbedtls/issues/1200 
I believe there will be more questions raised, if the the functions will not get the context.

AndrzejKurek:
@RonEld I've addressed your comments, please re-review.

authcrypt/authcrypt.h
```diff
@@ -23,6 +23,7 @@
 #include "mbedtls/cipher.h"
 #include "mbedtls/entropy.h"
 #include "mbedtls/ctr_drbg.h"
+#include "mbedtls/platform.h"
```
RonEld: In this case it's better to have `mbedtls_platform_context*` as forward declaration , instead of adding include for the header. I prefer minimal includes in header files, unless really needed, and keep the include in the `.cpp` file

AndrzejKurek:
Hi @RonEld, 
I've added the comment pointing to issue 1200. As for the two other requests:
- As much as I'd like to, I can't really forward declare an anonymous, typedef-ed struct. If you know a way to do it, please let me know.
- I'm against having a class member reference pointing to an externally managed object for the following reasons:

1. If the context gets deinitialized, there's nothing you can do with such dangling reference;
2. The reference can't be set to null if there's no context to be used;
3. They can't be easily replaced with smart pointers, if we ever decide to use any;
4. We're forced to initialize the reference in the constructor, and we can't change it later.

authcrypt/main.cpp
```diff
@@ -30,6 +30,8 @@ int main() {
     }
 
     int exit_code = MBEDTLS_EXIT_SUCCESS;
+    // The platform context is passed just in case any crypto calls need it.
+    // Please refer to https://github.com/ARMmbed/mbedtls/issues/1200 for more information.
```
RonEld: Please move this to the `Authcrypt` ctor, and have same comment like in the `hashing` and `benchmark` examples.

tls-client/main.cpp
```diff
@@ -443,6 +443,8 @@ int main() {
         return 1;
     }
 
+    // The platform context is passed just in case any crypto calls need it.
+    // Please refer to https://github.com/ARMmbed/mbedtls/issues/1200 for more information.
```
RonEld: Please move this to the `HelloHTTPS` ctor, and have same comment like in the `hashing` and `benchmark`

AndrzejKurek:
@RonEld - but the variable is being used, so the comment is irrelevant in the two classes that save it as a member.

RonEld:
You are right, but it is being used by setting a member of the class. We should be consistent, and have this member for `benchmarking` and `hash` as well.

AndrzejKurek:
I don't understand your request. Benchmark and hash tests don't use classes, they are just functions.

authcrypt/main.cpp
```diff
@@ -24,8 +24,13 @@
...
 
 int main() {
+    mbedtls_platform_context platform_ctx;
+    if(mbedtls_platform_setup(&platform_ctx) != 0) {
+        return -1;
+    }
+
     int exit_code = MBEDTLS_EXIT_SUCCESS;
-    Authcrypt *authcrypt = new Authcrypt();
+    Authcrypt *authcrypt = new Authcrypt(&platform_ctx);
```
andresag01: Question: I read the conversation about the location of these calls and the context. My understanding is that @RonEld suggested putting them closer to the crypto stuff. In the case of this particular example (and probably tls-client), why not make the struct part of the class itself?

hashing/main.cpp
```diff
@@ -152,8 +157,14 @@ static int example(void)
 }
 
 int main() {
-    int ret = example();
+    mbedtls_platform_context platform_ctx;
```
andresag01: Question: How large can these structs be? I think that the benchmark example used to have constant problems with stack size. Is this just going to contribute to making the problem worse?

tls-client/main.cpp
```diff
@@ -406,12 +409,17 @@ class HelloHTTPS {
...
     mbedtls_ssl_context _ssl;
     mbedtls_ssl_config _ssl_conf;
+    mbedtls_platform_context* _platform_ctx;
 };
 
 /**
  * The main loop of the HTTPS Hello World test
  */
 int main() {
+    mbedtls_platform_context platform_ctx;
```
andresag01: Same comment as with authcrypt. Please refer to my other comment.

hashing/main.cpp
andresag01: Agreed, but that is my point: that PR is not merged, so depending on the size of mbedtls_platform_context and what features are enabled by the config there might be stack issues.

sbutcher-arm:
@andresag01 

> could you please explain in a bit more detail what is the purpose of letting the user handle the platform setup/teardown steps?

Because that's what the HAL team asked us to do. To leave it to the user, and not place it in the boot sequence. It's their architecture, and their call, and it's not unreasonable.

authcrypt/main.cpp
AndrzejKurek: The platform initialization itself is not connected to a particular test starting. It's an external dependency which has to be satisfied. If someone was to split a test in two, or add another one, there should be only one platform initialization done independently of the test initialization.

hashing/main.cpp
RonEld: @andresag01 I pointed to that PR, to make a push on its merging:)

authcrypt/main.cpp
```diff
@@ -24,8 +24,13 @@
 #include "mbedtls/platform.h"
 
 int main() {
+    mbedtls_platform_context platform_ctx;
+    if(mbedtls_platform_setup(&platform_ctx) != 0) {
```
andresag01: I suggest not covering up the return code here and also adding something like:
```
printf( "Platform initialization failed with error -%04X",  -ret );
```
Not sure what the format should be but thats how we normally print error codes.

benchmark/main.cpp
```diff
@@ -961,8 +964,14 @@ static int benchmark( int argc, char *argv[] )
 }
 
 int main(void) {
-    int ret = benchmark(0, NULL);
+    mbedtls_platform_context platform_ctx;
+    if(mbedtls_platform_setup(&platform_ctx) != 0) {
+        return -1;
```
andresag01: I suggest not covering up the return code here and also adding something like:
```
printf( "Platform initialization failed with error -%04X",  -ret );
```
Not sure what the format should be but thats how we normally print error codes.

hashing/main.cpp
```diff
@@ -152,8 +157,14 @@ static int example(void)
...
+    if(mbedtls_platform_setup(&platform_ctx) != 0) {
```
andresag01: I suggest not covering up the return code here and also adding something like:
```
printf( "Platform initialization failed with error -%04X",  -ret );
```
Not sure what the format should be but thats how we normally print error codes.

tls-client/main.cpp
```diff
@@ -406,12 +409,17 @@ class HelloHTTPS {
...
+    if(mbedtls_platform_setup(&platform_ctx) != 0) {
+        return -1;
```
andresag01: I suggest not covering up the return code here and also adding something like:
```
printf( "Platform initialization failed with error -%04X",  -ret );
```
Not sure what the format should be but thats how we normally print error codes.

benchmark/main.cpp
```diff
@@ -961,8 +964,14 @@ static int benchmark( int argc, char *argv[] )
...
+    }
+
+    int ret = benchmark(0, NULL, &platform_ctx);
     if (ret != 0) {
         mbedtls_printf("Benchmark failed with error %d\r\n", ret);
     }
+    mbedtls_platform_teardown(&platform_ctx);
```
andresag01: Note that there are a lot of places inside `benchmark()` that call the function `mbedtls_exit()`. In these cases, the `mbedtls_platform_teardown()` will not be executed. Is this the intended behaviour?

sbutcher-arm:
@AndrzejKurek - Could you please look at @andresag01's review feedback?

AndrzejKurek:
Added and unified error printing for all tests, refactored the benchmark example to not exit unexpectedly.
@RonEld, @andresag01 - please re-review.

authcrypt/main.cpp
```diff
@@ -25,16 +25,18 @@
 
 int main() {
     mbedtls_platform_context platform_ctx;
-    if(mbedtls_platform_setup(&platform_ctx) != 0) {
-        return -1;
+    int exit_code = MBEDTLS_EXIT_FAILURE;
```
andresag01: As far as I can tell, regardless of whther mbedtls_platform_setup and run return, MBEDTLS_EXIT_FAILURE will always be returned. Is this correct?

benchmark/main.cpp
```diff
@@ -965,13 +965,19 @@ static int benchmark( int argc, char *argv[], mbedtls_platform_context* ctx )
 
 int main(void) {
     mbedtls_platform_context platform_ctx;
-    if(mbedtls_platform_setup(&platform_ctx) != 0) {
-        return -1;
+    int exit_code = MBEDTLS_EXIT_FAILURE;
```
andresag01: Again, I think this function will always return MBEDTLS_EXIT_FAILURE, a failure code...

hashing/main.cpp
```diff
@@ -158,13 +158,19 @@ static int example(mbedtls_platform_context* ctx)
 
 int main() {
     mbedtls_platform_context platform_ctx;
-    if(mbedtls_platform_setup(&platform_ctx) != 0) {
-        return -1;
+    int exit_code = MBEDTLS_EXIT_FAILURE;
```
andresag01: I think this function will always return MBEDTLS_EXIT_FAILURE, a failure code...

tls-client/main.cpp
```diff
@@ -417,8 +417,11 @@ class HelloHTTPS {
  */
 int main() {
     mbedtls_platform_context platform_ctx;
-    if(mbedtls_platform_setup(&platform_ctx) != 0) {
-        return -1;
+    int ret = MBEDTLS_EXIT_FAILURE;
```
andresag01: I think this function will always return MBEDTLS_EXIT_FAILURE, a failure code...

authcrypt/main.cpp
AndrzejKurek: No, the value will be overriden by platform_setup and authcrypt->run.

benchmark/main.cpp
AndrzejKurek: No, the value will be overriden by platform_setup and benchmark(...).

tls-client/main.cpp
AndrzejKurek: Ret is only used here to store the value of platform setup and print the error. When errors occur, sure, it returns MBEDTLS_EXIT_FAILURE, but when everything goes right, MBEDTLS_EXIT_SUCCESS is returned (at the end).

Patater:
We've just merged a large refactoring to the examples. In the interest of getting this platform PR in sooner, I've rebased and resolved conflicts. The old branch is available at https://github.com/ARMmbed/mbed-os-example-tls/tree/feature-platform-init-old for reference, if desired.

Patater:
CI passes on most targets, but got an out of memory error on KW24D with IAR.

```
14:13:08 [1525443188.04][CONN][RXD]   RSA-4096                 :  FAILED: RSA - The private key operation failed : BIGNUM - Memory allocation failed
```

AndrzejKurek:
@Patater - this error also happens on master.

authcrypt/authcrypt.cpp
```diff
@@ -41,8 +39,11 @@ const char Authcrypt::message[] = "Some things are better left unread";
 
 const char Authcrypt::metadata[] = "eg sequence number, routing info";
 
-Authcrypt::Authcrypt()
+Authcrypt::Authcrypt(mbedtls_platform_context* platform_ctx)
```
mazimkhan: Why is it required to pass ```platform_ctx``` to the application(test)?

benchmark/main.cpp
```diff
@@ -300,7 +300,7 @@ typedef struct {
          rsa, dhm, ecdsa, ecdh;
 } todo_list;
 
-static int benchmark( int argc, char *argv[] )
+static int benchmark( int argc, char *argv[], mbedtls_platform_context* ctx )
```
mazimkhan: Same question as https://github.com/ARMmbed/mbed-os-example-tls/pull/165/files#r186700998

benchmark/main.cpp
```diff
@@ -309,7 +309,10 @@ static int benchmark( int argc, char *argv[] )
 #if defined(MBEDTLS_MEMORY_BUFFER_ALLOC_C)
     unsigned char malloc_buf[HEAP_SIZE] = { 0 };
 #endif
-
+    // The call below is used to avoid the "unused parameter" warning.
+    // The context itself can be used by cryptographic calls which require it.
```
mazimkhan: An application like this test will not directly call a cryptographic function. The cryptographic functions should have a way of accessing the context like a ```getter``` as suggested in https://github.com/ARMmbed/mbedtls/issues/1200

hashing/main.cpp
```diff
@@ -152,8 +157,20 @@ static int example(void)
...
 
 int main() {
-    int ret = example();
-    if (ret != 0) {
-        mbedtls_printf("Example failed with error %d\r\n", ret);
+    mbedtls_platform_context platform_ctx;
+    int exit_code = MBEDTLS_EXIT_FAILURE;
+
+    if((exit_code = mbedtls_platform_setup(&platform_ctx)) != 0) {
+        printf("Platform initialization failed with error %d\r\n", exit_code);
```
mazimkhan: ```printf``` -> ```mbedtls_printf```

authcrypt/authcrypt.cpp
mazimkhan: This can be limited to the tests that actually test things from the platform context.

hashing/main.cpp
mazimkhan: Agreed.

PR #173, opened 2018-06-04
easy-connect v1.2.10

JanneKiiskila:
Bring in Nucleo-WiFi-X fixes.


JanneKiiskila:
These aren't moving ahead fast enough, we have a new version already.

PR #174, opened 2018-06-05
Separate server name

RonEld:
Seperate the server_name into the `server_addr`
which is the address to connect,
and to `server_name` which is the server host_name.
Sometimes these are not the same, and it helps separating them for connecting to other servers

### DIFF
--- a/tls-client/HelloHttpsClient.cpp
+++ b/tls-client/HelloHttpsClient.cpp
@@ -81,7 +81,6 @@ HelloHttpsClient::HelloHttpsClient(const char *in_server_name,
      * Please refer to https://github.com/ARMmbed/mbedtls/issues/1200 for more
      * information. */
     platform_ctx(in_platform_ctx)
-
 {
     mbedtls_entropy_init(&entropy);
     mbedtls_ctr_drbg_init(&ctr_drbg);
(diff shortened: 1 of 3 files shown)


tls-client/HelloHttpsClient.cpp
```diff
@@ -70,15 +70,18 @@ const char *HelloHttpsClient::HTTP_HELLO_STR = "Hello world!";
...
                                    mbedtls_platform_context* in_platform_ctx) :
     socket(),
     server_name(in_server_name),
+    server_addr(in_server_addr),
     server_port(in_server_port),
     /* The platform context is passed just in case any crypto calls need it.
      * Please refer to https://github.com/ARMmbed/mbedtls/issues/1200 for more
      * information. */
     platform_ctx(in_platform_ctx)
+
```
andresag01: Could you please remove this whitespace?

RonEld: ok

RonEld:
@andresag01 I agree that the examples should be simple, 
I have encountered several times that the server addr is different than the server name in the CN of the server's certificate. This application is used for reference and POC with different servers.
For example, if you use the exampleMbed TLS server on your PC (ssl_server2), and you want to connect your coard to your server, you will need a different server address ( your PC's IP \ computer name ) and server Name ( the CN of the default test certificate - localhost)
I have managed to connect to real servers ( MQTT servers in my case ), where the certificate CN was different than the server address.
If the hostname isn't equal to the Subject CN of the certificate, then certificate verification fails

RonEld:
@andresag01 I remove the extra whitespace which was accidentally introduced in a merge conflict resolution

andresag01:
All the relevant tests in the CI pass.

PR #186, opened 2018-07-30
use mbedtls_printf instead of printf

### LINKED ISSUE #185: benchmark application uses printf instead of mbedtls_printf
[benchmark.c](https://github.com/ARMmbed/mbed-os-example-tls/blame/master/benchmark/main.cpp#L1022) contains `printf()` instead of the abstraction`mbedtls_printf()`

RonEld:
Change call to `mbedtls_printf()` instead of`printf()`
which was accidently inserted.
Resolves #185


RonEld:
closing, as this is covered by #194

PR #189, opened 2018-08-01
Revert addition of ecdsa test

### LINKED ISSUE #184: Benchmark example fails on K64F with ARM C Compiler
The Benchmark example is failing at the ECDSA test (again) on K64F and NUCLEO_F429ZI with the ARM C Compiler.

This is a regression introduced by PR #116. 

We need to revert this change.

cc: @k-stachowiak, @RonEld

RonEld:
Revert the addition of ecdsa test, as it seems that CI still fails.
Resolves #184.

### DIFF
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -145,7 +145,7 @@
 /*
  * Uncomment this line to enable ECDSA benchmark.
  */
-#define ENABLE_ECDSA
+//#define ENABLE_ECDSA
 
 /*
  * For heap usage estimates, we need an estimate of the overhead per allocated


andresag01:
Note that https://github.com/ARMmbed/mbed-os-example-tls/pull/194 reworks the benchmark example to address this problem (among others)

AndrzejKurek:
I would also opt for Andres' solution, than for disabling the tests.

RonEld:
I agree, and when I removed the ECDSA disabling, the tests worked for me, however ,in CI, the tests failed for the same board

sbutcher-arm:
retest

RonEld:
@sbutcher-arm Do we need this merged, assuming #194 will fix the memory issues?

sbutcher-arm:
PR #194 fixes the problem properly, so I don't think we need this PR now. Thanks for supplying it anyway @RonEld.

PR #190, opened 2018-08-01
Remove MBEDTLS_MEMORY_BUFFER_ALLOC_C usages

### LINKED ISSUE #171: MBEDTLS_MEMORY_BUFFER_ALLOC_C feature in benchmark is problematic
The `MBEDTLS_MEMORY_BUFFER_ALLOC_C` feature in the `benchmark` application is problematic, as it uses 64 KB of stack size. By default Mbed-OS uses 4 KB of stack size. This feature should either be removed, or length should be taken from the link script

RonEld:
Remove references to `MBEDTLS_MEMORY_BUFFER_ALLOC_C` as it is not used
in Mbed OS, and Mbed OS has its own memory statistics mechanisms.
Resolves #171

### DIFF
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -143,13 +143,6 @@
  */
 #define ENABLE_ECDSA
 
-/*
- * For heap usage estimates, we need an estimate of the overhead per allocated
- * block. ptmalloc2/3 (used in gnu libc for instance) uses 2 size_t per block,
- * so use that as our baseline.
- */
-#define MEM_BLOCK_OVERHEAD  ( 2 * sizeof( size_t ) )
-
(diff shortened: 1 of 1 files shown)


benchmark/main.cpp
```diff
@@ -154,11 +150,6 @@
  */
 #define MEM_BLOCK_OVERHEAD  ( 2 * sizeof( size_t ) )
```
sbutcher-arm: This line can also, and therefore should also be removed.

RonEld:
@sbutcher-arm I addressed your comment

RonEld:
Closing, as #194 covers this as well

PR #193, opened 2018-08-13
easy-connect v1.2.16

JanneKiiskila:
Updated drivers.
Eliminate static data when unreferenced.

### DIFF
--- a/tls-client/easy-connect.lib
+++ b/tls-client/easy-connect.lib
@@ -1,1 +1,1 @@
-https://github.com/ARMmbed/easy-connect/#5b52b5fa56c623d5853c5daf266b34986a0c20cc
+https://github.com/ARMmbed/easy-connect/#bf48f5b649eba3ed7b810c77c630b5eb51301311


JanneKiiskila:
Please review  - @RonEld  and  andresag01

tls-client/easy-connect.lib
```diff
@@ -1 +1 @@
-https://github.com/ARMmbed/easy-connect/#5b52b5fa56c623d5853c5daf266b34986a0c20cc
+https://github.com/ARMmbed/easy-connect/#bf48f5b649eba3ed7b810c77c630b5eb51301311
```
RonEld: Wouldn't it work if we change to https://github.com/ARMmbed/easy-connect/tree/v1.2.16 ?
I think it will be more readable this way

JanneKiiskila:
Would be nice if it did, but I understood `mbed cli` toolchain does not like that? @screamerbg  - do we support that syntax?

RonEld:
I tried it, and unfortunately it didn't work

JanneKiiskila:
But, I fully agree with you - it would be so much nicer to just read the version number rather than some cryptic hash!

andresag01:
@JanneKiiskila: Thank you for the PR. The PR looks good to me. I have now started a CI run and will wait for the results before approving.

JanneKiiskila:
Seems to fail with KW24D.

```
[build-benchmark-KW24D-IAR] [1534195522.07][CONN][RXD]   RSA-4096                 :  FAILED: RSA - The private key operation failed : BIGNUM - Memory allocation failed
```

But, well, this one should not impact that in any way.

sbutcher-arm:
CI is failing due to memory allocation - a known issue. Therefore labelling as 'ready to merge'.

RonEld:
note: Since Mbed OS 5.9, we should be using NetworkInterface
see #198 for more information.
easy-connect doesn't compile for some platforms, (I tested on NRF52840_DK) due to this latest change.
I think we should close this PR, and merge #198 , hence marking this PR as changes requires

JanneKiiskila:
Upto 5.9.x releases we should still use easy-connect, as the default network interface does not work for anything BUT ethernet. But, yes starting with 5.10 we should move to the proper interface.

tls-client/easy-connect.lib
JanneKiiskila: Mbed cli tooling does not support that, unfortunately.

JanneKiiskila:
That's sort of not possible, because the easy-connect pulls in some repos and after 5.10, Mbed OS will also pull in some (but not all!) of the same repos -> code duplication and you need either to delete easy-connect.lib OR have a Mbed OS version specific .mbedignore -file. No easy way out here.

RonEld:
I agree.
I think the only easy way out of this, is to merge this PR, create a branch \ tag , and then remove easy-connect.
The downside is backporting bug fixes, but we can decide this will be a version tag, and no bug backporting neccessary

RonEld:
I got build error for NRF52480_DK platform, because of the usage of the NetworkInterface:
~~~
BUILD/NRF52840_DK/GCC_ARM/easy-connect/easy-connect.o: In function `__static_initialization_and_destruction_0':
/home/roneld01/work/mbed-os-example-tls/tls-client/./mbed-os/features/netsocket/EthernetInterface.h:45: undefined reference to `EMAC::get_default_instance()'
/home/roneld01/work/mbed-os-example-tls/tls-client/./mbed-os/features/netsocket/EthernetInterface.h:46: undefined reference to `OnboardNetworkStack::get_default_instance()'
collect2: error: ld returned 1 exit status
[ERROR] BUILD/NRF52840_DK/GCC_ARM/easy-connect/easy-connect.o: In function `__static_initialization_and_destruction_0':
/home/roneld01/work/mbed-os-example-tls/tls-client/./mbed-os/features/netsocket/EthernetInterface.h:45: undefined reference to `EMAC::get_default_instance()'
/home/roneld01/work/mbed-os-example-tls/tls-client/./mbed-os/features/netsocket/EthernetInterface.h:46: undefined reference to `OnboardNetworkStack::get_default_instance()'
collect2: error: ld returned 1 exit status
~~~

Since we don't have maintained legacy branches, we should probably close this PR, IMHO. Are there significant bug fixes that will require us to create a branch \ tag for an earlier version of Mbed OS with this PR?

JanneKiiskila:
No bugs that I am aware. However, the example compilation problem with NRF is not related to the easy-connect repo - that's just a matter of fact, if you try to use the default network interface with a board, that does not have one (like this NRF board). That same compilation problem will happen also without easy-connect, unless you configure it properly.

Issue #172, opened 2018-05-23
benchmark application: TIME_AND_TSC doesn't check return code


RonEld:
The return code of function calls in `TIME_AND_TSC` is not checked

ciarmcom:
ARM Internal Ref: IOTSSL-2418

andresag01:
@sbutcher-arm, @RonEld: I think this was fixed by https://github.com/ARMmbed/mbed-os-example-tls/pull/194. Should we close this issue

andresag01:
This issue has been fixed via #194, so can now be closed.

PR #194, opened 2018-08-13
Refactor benchmark to mitigate memory problems

andresag01:
This PR entirely refactors the benchmark example to fix the memory usage issues in Mbed OS (particularly stack overflows). The idea is to run the benchmark for each module in a completely different (noinline) function. Also, some objects are moved to either to the global section or are allocated dynamically.

This PR also includes changes to:
* Remove \r\n in favour of \n
* Remove the todo struct which was a remainder from when the program was command line (https://github.com/ARMmbed/mbed-os-example-tls/issues/170)
* Check for errors at every stage (https://github.com/ARMmbed/mbed-os-example-tls/issues/172)
* Enable more test cases by manually defining more MBEDTLS_* macros in a user config
* Consistently use mbedtls_printf (https://github.com/ARMmbed/mbed-os-example-tls/issues/185)
* Remove use of memory_buffer_alloc module (https://github.com/ARMmbed/mbed-os-example-tls/issues/171)

Also addresses: https://github.com/ARMmbed/mbed-os-example-tls/issues/52, https://github.com/ARMmbed/mbed-os-example-tls/issues/29, https://github.com/ARMmbed/mbed-os-example-tls/issues/15 and https://github.com/ARMmbed/mbed-os-example-tls/issues/184

**NOTE:** This PR is currently under development. I have checked that it works in K64F, but I am waiting for the CI to complete to check whether any other memory issues are detected in the other targets. Regardless, the main structure of the change is in place and can be reviewed.

### DIFF
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -301,9 +301,10 @@ exit:
 }
 #endif /* MBEDTLS_SHA512_C */
 
+
+#if defined(MBEDTLS_ARC4_C)
 MBED_NOINLINE static int benchmark_arc4() {
     int ret = 0;
-#if defined(MBEDTLS_ARC4_C)
     mbedtls_arc4_context arc4;
 
     mbedtls_arc4_init(&arc4);
@@ -314,14 +315,14 @@ MBED_NOINLINE static int benchmark_arc4() {
 
 exit:
     mbedtls_arc4_free(&arc4);
-#endif /* MBEDTLS_ARC4_C */
 
     return ret;
 }
+#endif /* MBEDTLS_ARC4_C */
 
+#if defined(MBEDTLS_DES_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
 MBED_NOINLINE static int benchmark_des3() {
     int ret = 0;
-#if defined(MBEDTLS_DES_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
     mbedtls_des3_context des3;
 
     mbedtls_des3_init(&des3);
@@ -336,14 +337,14 @@ MBED_NOINLINE static int benchmark_des3() {
 
 exit:
     mbedtls_des3_free(&des3);
-#endif /* MBEDTLS_DES_C && MBEDTLS_CIPHER_MODE_CBC */
 
     return ret;
 }
+#endif /* MBEDTLS_DES_C && MBEDTLS_CIPHER_MODE_CBC */
 
+#if defined(MBEDTLS_DES_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
 MBED_NOINLINE static int benchmark_des() {
     int ret = 0;
-#if defined(MBEDTLS_DES_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
     mbedtls_des_context des;
 
     mbedtls_des_init(&des);
@@ -358,15 +359,15 @@ MBED_NOINLINE static int benchmark_des() {
 
 exit:
     mbedtls_des_free(&des);
-#endif /* MBEDTLS_DES_C && MBEDTLS_CIPHER_MODE_CBC */
 
     return ret;
 }
+#endif /* MBEDTLS_DES_C && MBEDTLS_CIPHER_MODE_CBC */
 
-MBED_NOINLINE static int benchmark_des3_cmac() {
-    int ret = 0;
 #if defined(MBEDTLS_DES_C) && defined(MBEDTLS_CIPHER_MODE_CBC) && \
     defined(MBEDTLS_CMAC_C)
+MBED_NOINLINE static int benchmark_des3_cmac() {
+    int ret = 0;
     unsigned char output[8];
     const mbedtls_cipher_info_t *cipher_info;
 
@@ -384,13 +385,13 @@ MBED_NOINLINE static int benchmark_des3_cmac() {
                                             BUFSIZE, output));
 
 exit:
-#endif /* MBEDTLS_DES_C && MBEDTLS_CIPHER_MODE_CBC && MBEDTLS_CMAC_C */
     return ret;
 }
+#endif /* MBEDTLS_DES_C && MBEDTLS_CIPHER_MODE_CBC && MBEDTLS_CMAC_C */
 
+#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
 MBED_NOINLINE static int benchmark_aes_cbc() {
     int ret = 0;
-#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
     int keysize;
     mbedtls_aes_context aes;
 
@@ -421,14 +422,14 @@ MBED_NOINLINE static int benchmark_aes_cbc() {
 
 exit:
     mbedtls_aes_free(&aes);
-#endif /* MBEDTLS_AES_C && MBEDTLS_CIPHER_MODE_CBC */
 
     return ret;
 }
+#endif /* MBEDTLS_AES_C && MBEDTLS_CIPHER_MODE_CBC */
 
+#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CIPHER_MODE_CTR)
 MBED_NOINLINE static int benchmark_aes_ctr() {
     int ret = 0;
-#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CIPHER_MODE_CTR)
     int keysize;
     size_t nc_offset = 0;
     unsigned char stream_block[16];
@@ -461,14 +462,14 @@ MBED_NOINLINE static int benchmark_aes_ctr() {
 
 exit:
     mbedtls_aes_free(&aes);
-#endif /* MBEDTLS_AES_C && MBEDTLS_CIPHER_MODE_CTR */
 
     return ret;
 }
+#endif /* MBEDTLS_AES_C && MBEDTLS_CIPHER_MODE_CTR */
 
+#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_GCM_C)
 MBED_NOINLINE static int benchmark_aes_gcm() {
     int ret = 0;
-#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_GCM_C)
     int keysize;
     mbedtls_gcm_context gcm;
 
@@ -500,14 +501,14 @@ MBED_NOINLINE static int benchmark_aes_gcm() {
 
 exit:
     mbedtls_gcm_free(&gcm);
-#endif /* MBEDTLS_AES_C && MBEDTLS_GCM_C */
 
     return ret;
 }
+#endif /* MBEDTLS_AES_C && MBEDTLS_GCM_C */
 
+#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CCM_C)
 MBED_NOINLINE static int benchmark_aes_ccm() {
     int ret = 0;
-#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CCM_C)
     int keysize;
     mbedtls_ccm_context ccm;
 
@@ -538,14 +539,14 @@ MBED_NOINLINE static int benchmark_aes_ccm() {
 
 exit:
     mbedtls_ccm_free(&ccm);
-#endif /* MBEDTLS_AES_C && MBEDTLS_CCM_C */
 
     return ret;
 }
+#endif /* MBEDTLS_AES_C && MBEDTLS_CCM_C */
 
+#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CMAC_C)
 MBED_NOINLINE static int benchmark_aes_cmac() {
     int ret = 0;
-#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CMAC_C)
     unsigned char output[16];
     const mbedtls_cipher_info_t *cipher_info;
     mbedtls_cipher_type_t cipher_type;
@@ -583,14 +584,14 @@ MBED_NOINLINE static int benchmark_aes_cmac() {
                         output));
 
 exit:
-#endif /* MBEDTLS_AES_C && MBEDTLS_CMAC_C */
 
     return ret;
 }
+#endif /* MBEDTLS_AES_C && MBEDTLS_CMAC_C */
 
+#if defined(MBEDTLS_CAMELLIA_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
 MBED_NOINLINE static int benchmark_camellia() {
     int ret = 0;
-#if defined(MBEDTLS_CAMELLIA_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
     int keysize;
     mbedtls_camellia_context camellia;
 
@@ -623,14 +624,14 @@ MBED_NOINLINE static int benchmark_camellia() {
 
 exit:
         mbedtls_camellia_free(&camellia);
-#endif /* MBEDTLS_CAMELLIA_C && MBEDTLS_CIPHER_MODE_CBC */
 
     return ret;
 }
+#endif /* MBEDTLS_CAMELLIA_C && MBEDTLS_CIPHER_MODE_CBC */
 
+#if defined(MBEDTLS_BLOWFISH_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
 MBED_NOINLINE static int benchmark_blowfish() {
     int ret = 0;
-#if defined(MBEDTLS_BLOWFISH_C) && defined(MBEDTLS_CIPHER_MODE_CBC)
     int keysize;
     mbedtls_blowfish_context *blowfish;
 
@@ -670,14 +671,14 @@ MBED_NOINLINE static int benchmark_blowfish() {
 exit:
     mbedtls_blowfish_free(blowfish);
     mbedtls_free(blowfish);
-#endif /* MBEDTLS_BLOWFISH_C && MBEDTLS_CIPHER_MODE_CBC */
 
     return ret;
 }
+#endif /* MBEDTLS_BLOWFISH_C && MBEDTLS_CIPHER_MODE_CBC */
 
+#if defined(MBEDTLS_HAVEGE_C)
 MBED_NOINLINE static int benchmark_havege() {
     int ret = 0;
-#if defined(MBEDTLS_HAVEGE_C)
     mbedtls_havege_state hs;
 
     mbedtls_havege_init(&hs);
@@ -686,14 +687,14 @@ MBED_NOINLINE static int benchmark_havege() {
 
 exit:
     mbedtls_havege_free(&hs);
-#endif /* MBEDTLS_HAVEGE_C */
 
     return ret;
 }
+#endif /* MBEDTLS_HAVEGE_C */
 
+#if defined(MBEDTLS_CTR_DRBG_C)
 MBED_NOINLINE static int benchmark_ctr_drbg() {
     int ret = 0;
-#if defined(MBEDTLS_CTR_DRBG_C)
     mbedtls_ctr_drbg_context ctr_drbg;
 
     mbedtls_ctr_drbg_init(&ctr_drbg);
@@ -720,14 +721,14 @@ MBED_NOINLINE static int benchmark_ctr_drbg() {
 
 exit:
     mbedtls_ctr_drbg_free( &ctr_drbg );
-#endif /* MBEDTLS_CTR_DRBG_C */
 
     return ret;
 }
+#endif /* MBEDTLS_CTR_DRBG_C */
 
+#if defined(MBEDTLS_HMAC_DRBG_C)
 MBED_NOINLINE static int benchmark_hmac_drbg() {
     int ret = 0;
-#if defined(MBEDTLS_HMAC_DRBG_C)
     mbedtls_hmac_drbg_context hmac_drbg;
     const mbedtls_md_info_t *md_info;
 
@@ -788,15 +789,15 @@ MBED_NOINLINE static int benchmark_hmac_drbg() {
 
 exit:
     mbedtls_hmac_drbg_free(&hmac_drbg);
-#endif /* MBEDTLS_HMAC_DRBG_C */
 
     return ret;
 }
+#endif /* MBEDTLS_HMAC_DRBG_C */
 
-MBED_NOINLINE static int benchmark_rsa() {
-    int ret = 0;
 #if defined(MBEDTLS_RSA_C) && \
     defined(MBEDTLS_PEM_PARSE_C) && defined(MBEDTLS_PK_PARSE_C)
+MBED_NOINLINE static int benchmark_rsa() {
+    int ret = 0;
     mbedtls_pk_context pk;
     mbedtls_rsa_context *rsa;
     const char *rsa_keys[] = {
@@ -837,14 +838,14 @@ MBED_NOINLINE static int benchmark_rsa() {
 exit:
         mbedtls_pk_free(&pk);
     }
-#endif /* MBEDTLS_RSA_C && MBEDTLS_PEM_PARSE_C && MBEDTLS_PK_PARSE_C */
 
     return ret;
 }
+#endif /* MBEDTLS_RSA_C && MBEDTLS_PEM_PARSE_C && MBEDTLS_PK_PARSE_C */
 
+#if defined(MBEDTLS_DHM_C) && defined(MBEDTLS_BIGNUM_C)
 MBED_NOINLINE static int benchmark_dhm() {
     int ret = 0;
-#if defined(MBEDTLS_DHM_C) && defined(MBEDTLS_BIGNUM_C)
     int dhm_sizes[] = {
         2048,
         3072,
@@ -923,14 +924,14 @@ MBED_NOINLINE static int benchmark_dhm() {
 exit:
         mbedtls_dhm_free(&dhm);
     }
-#endif /* MBEDTLS_DHM_C && MBEDTLS_BIGNUM_C */
 
     return ret;
 }
+#endif /* MBEDTLS_DHM_C && MBEDTLS_BIGNUM_C */
 
+#if defined(MBEDTLS_ECDSA_C) && defined(MBEDTLS_SHA256_C)
 MBED_NOINLINE static int benchmark_ecdsa() {
     int ret = 0;
-#if defined(MBEDTLS_ECDSA_C) && defined(MBEDTLS_SHA256_C)
     mbedtls_ecdsa_context ecdsa;
     const mbedtls_ecp_curve_info *curve_info;
     size_t sig_len;
@@ -1009,15 +1010,14 @@ MBED_NOINLINE static int benchmark_ecdsa() {
 exit:
         mbedtls_ecdsa_free(&ecdsa);
     }
-#endif /* MBEDTLS_ECDSA_C && MBEDTLS_SHA2565_C */
-
 
     return ret;
 }
+#endif /* MBEDTLS_ECDSA_C && MBEDTLS_SHA2565_C */
 
+#if defined(MBEDTLS_ECDH_C)
 MBED_NOINLINE static int benchmark_ecdh() {
     int ret = 0;
-#if defined(MBEDTLS_ECDH_C)
     mbedtls_ecdh_context ecdh;
     const mbedtls_ecp_curve_info *curve_info;
     size_t olen;
@@ -1119,15 +1119,15 @@ MBED_NOINLINE static int benchmark_ecdh() {
 exit:
         mbedtls_ecdh_free(&ecdh);
     }
-#endif /* MBEDTLS_ECDH_C */
 
     return ret;
 }
+#endif /* MBEDTLS_ECDH_C */
 
+#if defined(MBEDTLS_ECDH_C) && defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED)
 /* Curve25519 needs to be handled separately */
 MBED_NOINLINE static int benchmark_ecdh_curve22519() {
     int ret = 0;
-#if defined(MBEDTLS_ECDH_C) && defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED)
     mbedtls_ecdh_context ecdh;
     mbedtls_mpi z;
 
@@ -1188,16 +1188,15 @@ MBED_NOINLINE static int benchmark_ecdh_curve22519() {
 exit:
     mbedtls_ecdh_free(&ecdh);
     mbedtls_mpi_free(&z);
-#endif /* MBEDTLS_ECDH_C && MBEDTLS_ECP_DP_CURVE25519_ENABLED */
 
     return ret;
 }
+#endif /* MBEDTLS_ECDH_C && MBEDTLS_ECP_DP_CURVE25519_ENABLED */
 
 int main()
 {
     mbedtls_platform_context platform_ctx;
-    int exit_code = MBEDTLS_EXIT_FAILURE;
-    int ret;
+    int exit_code = MBEDTLS_EXIT_SUCCESS;
 
     memset(buf, 0xAA, sizeof(buf));
     memset(tmp, 0xBB, sizeof(tmp));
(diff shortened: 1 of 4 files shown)


RonEld:
@andresag01 Before I do a full review, I have a question and a couple of comments:
- #116 Should have handled most of the memory usage issue. Wasn't this the case
- The todo struct is already removed in #170 
- The usage of the memory_buffer_alloc is already removed in #190 

Since these last two PRs already exist, I think it would be much simpler to review if you revert your changes related to the todo struct and memory_buffer_alloc

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
-#if defined(MBEDTLS_ERROR_C)
-#define PRINT_ERROR                                            \
-        mbedtls_strerror( ret, ( char * )tmp, sizeof( tmp ) ); \
-        mbedtls_printf( "FAILED: %s\r\n", tmp );
+#define BENCHMARK_FUNC_CALL(TITLE, CODE)                                    \
+do {                                                                        \
+    unsigned long i;                                                        \
+    Timeout t;                                                              \
+                                                                            \
+    mbedtls_printf(HEADER_FORMAT, TITLE);                                   \
```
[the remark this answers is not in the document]
andresag01: It is not because I am trying to use the Mbed OS style as this is an Mbed OS example

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
+        goto exit;                                      \
+    } else {                                            \
+        mbedtls_printf("%6lu ms/" TYPE, ms);            \
+        mbedtls_printf("\r\n");                         \
+    }                                                   \
+} while(0)
+
+/* Clear some memory that was used to prepare the context */
+#if defined(MBEDTLS_ECP_C)
+void ecp_clear_precomputed(mbedtls_ecp_group *grp) {
```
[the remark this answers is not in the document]
andresag01: Correct, I am trying to use the Mbed OS style, but missed a lot of cases. I will just run astyle to format everything automatically

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
-    ms = t.read_ms();                               \
-                                                    \
-    if( ret != 0 )                                  \
-    {                                               \
-        PRINT_ERROR;                                \
-    }                                               \
-    else                                            \
-    {                                               \
-        mbedtls_printf( "%6lu ms/" TYPE, ms );      \
-        MEMORY_MEASURE_PRINT( sizeof( TYPE ) + 1 ); \
```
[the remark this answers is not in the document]
andresag01: MEMORY_MEASURE_PRINT is dependent on memory_buffer_alloc. However, the way it was used here causes memory issues, so I am removing it. There is also a ticket to do that...

andresag01:
@RonEld:
>#116 Should have handled most of the memory usage issue. Wasn't this the case

I think that #116 was a good step, but looking at the assembly source for GCC the stack allocation was still very high. Also, I investigated further and some primitive overflow the stack even if they are the only benchmark in the function. So I decided to split everything and tag functions with noinline in the hope that this takes care of the problem once and for all. Also, I moved one or two of the most memory consuming contexts to dynamic allocation as it would not work otherwise.

>* The todo struct is already removed in #170
* The usage of the memory_buffer_alloc is already removed in #190

My original plan was to just fix the worst memory issues, but then I ended up making so many changes that I just decided to include fixes for those two issues as well here. In fact, this PR is essentially a rewrite of the application rather than a "fix". I included changes for issues such as todo and memory_buffer_alloc because it was simple to do it here and would be more time consuming if we had to resolve conflicts twice across three PRs.

RonEld:
@andresag01 
> Also, I investigated further and some primitive overflow the stack even if they are the only benchmark in the function.
Yes, https://github.com/ARMmbed/mbedtls/pull/1937 should resolve the overflow in the SHA functions

andresag01:
@RonEld: Actually I did not have an issue with the hash primitives. However, something like Blowfish is more problematic.

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
+        if (ret < 0 || static_cast<size_t>(ret) >= sizeof(title)) {
+            mbedtls_printf("Failed to compose title string using "
+                           "mbedtls_snprintf(): %d\n", ret);
+            goto exit;
+        }
+
+        memset(buf, 0, sizeof(buf));
+        memset(tmp, 0, sizeof(tmp));
+
+        if ((ret = mbedtls_aes_setkey_enc(&aes, tmp, keysize)) != 0) {
```
RonEld: Please skip in case of `MBEDTLS_AES_FEATURE_NOT_SUPPORTED`

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
-            TIME_AND_TSC( title,
-                mbedtls_aes_crypt_ctr( &aes, BUFSIZE, &nc_offset, tmp, stream_block, buf, buf ) );
+exit:
+    mbedtls_aes_free(&aes);
+#endif /* MBEDTLS_AES_C && MBEDTLS_CIPHER_MODE_CBC */
+
+    return ret;
+}
+
+MBED_NOINLINE static int benchmark_aes_ctr() {
```
RonEld: why not surround the whole function with `#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CIPHER_MODE_CTR)`

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
+            mbedtls_printf("Failed to compose title string using "
+                           "mbedtls_snprintf(): %d\n", ret);
+            goto exit;
         }
-        mbedtls_aes_free( &aes );
+
+        memset(buf, 0, sizeof(buf));
+        memset(tmp, 0, sizeof(tmp));
+
+        if ((ret = mbedtls_aes_setkey_enc(&aes, tmp, keysize)) != 0) {
```
RonEld: Please skip in case of feature not supported

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
-        int keysize;
-        mbedtls_gcm_context gcm;
+exit:
+    mbedtls_aes_free(&aes);
+#endif /* MBEDTLS_AES_C && MBEDTLS_CIPHER_MODE_CTR */
+
+    return ret;
+}
+
+MBED_NOINLINE static int benchmark_aes_gcm() {
```
RonEld: why not surround the whole function with `#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_GCM_C)`?

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
-            TIME_AND_TSC( title,
-                    mbedtls_gcm_crypt_and_tag( &gcm, MBEDTLS_GCM_ENCRYPT, BUFSIZE, tmp,
-                        12, NULL, 0, buf, buf, 16, tmp ) );
+        memset(buf, 0, sizeof(buf));
+        memset(tmp, 0, sizeof(tmp));
 
-            mbedtls_gcm_free( &gcm );
+        ret = mbedtls_gcm_setkey(&gcm, MBEDTLS_CIPHER_ID_AES, tmp, keysize);
+        if (ret != 0) {
+            mbedtls_printf("mbedtls_gcm_setkey() returned -0x%04X\n", -ret);
```
RonEld: Please skip in case of feature not supported

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
+exit:
+    mbedtls_gcm_free(&gcm);
+#endif /* MBEDTLS_AES_C && MBEDTLS_GCM_C */
+
+    return ret;
+}
+
+MBED_NOINLINE static int benchmark_aes_ccm() {
+    int ret = 0;
+#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CCM_C)
```
RonEld: why not surround the whole function with `#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CCM_C)`?

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
+exit:
+    mbedtls_ccm_free(&ccm);
+#endif /* MBEDTLS_AES_C && MBEDTLS_CCM_C */
+
+    return ret;
+}
+
+MBED_NOINLINE static int benchmark_aes_cmac() {
+    int ret = 0;
+#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CMAC_C)
```
RonEld: why not surround the whole function with `#if defined(MBEDTLS_AES_C) && defined(MBEDTLS_CMAC_C)`?

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
+                        mbedtls_aes_cmac_prf_128(tmp, 16, buf, BUFSIZE,
+                        output));
+
+exit:
+#endif /* MBEDTLS_AES_C && MBEDTLS_CMAC_C */
+
+    return ret;
+}
+
+MBED_NOINLINE static int benchmark_camellia() {
```
RonEld: why not surround the whole function with `#if defined(MBEDTLS_CAMELLIA_C) && defined(MBEDTLS_CIPHER_MODE_CBC)`?

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
-            mbedtls_ecdh_gen_public( &ecdh.grp, &ecdh.d, &ecdh.Qp, myrand, NULL ) != 0 )
-        {
-            return( 1 );
+        ret =mbedtls_ecp_copy( &ecdh.Qp, &ecdh.Q );
+        if (ret != 0) {
+            mbedtls_printf("mbedtls_ecp_copy() returned -0x%04X\n", -ret);
+            goto exit;
         }
 
-        TIME_PUBLIC(  "ECDHE-Curve25519"
```
RonEld: I know this is the original behaviour, but if we are alreadey doing a rewrite, please don't exit on failure, we should continue to next test.

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
-    }
-    grp->T = NULL;
-    grp->T_size = 0;
-}
-#else
-#define ecp_clear_precomputed( g )
-#endif
+MBED_NOINLINE static int benchmark_arc4() {
+    int ret = 0;
+#if defined(MBEDTLS_ARC4_C)
```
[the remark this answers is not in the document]
andresag01: Because then I would have to duplicate the preprocessor check here and in the main() function.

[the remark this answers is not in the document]
andresag01: I think code size here is not really a problem. When the feature is not defined as a macro then the compiler (GCC) will generate the following code for an empty function:
```
0000181c <benchmark_havege()>:
    181c:	2000      	movs	r0, #0
    181e:	4770      	bx	lr
```
Then there will be the following call in the main:
```
    36aa:	f7fe f8b7 	bl	181c <benchmark_havege()>
    36ae:	b9b8      	cbnz	r0, 36e0 <main+0x37c>
```
Also, I checked and when multiple functions are empty, then the compiler will reuse one of the empty functions instead of emitting multiple ones to reduce code size. In any case, the overhead is 2.5 words (10 bytes) which take about 8-10 clock cycles to execute in a Cortex-M0...

benchmark/main.cpp
andresag01: I think AES has the feature unavailable macro error code, but as far as I know GCM does not have it yet. I think the PR that adds it is: https://github.com/ARMmbed/mbedtls/pull/1716. Is this correct? If so, then I will not add the check because the symbol MBEDTLS_ERR_GCM_FEATURE_UNAVAILABLE does not exist yet. Or is it possible that this function returns MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE?

benchmark/main.cpp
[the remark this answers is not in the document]
RonEld: Have you tried all toolchains? ARM, GCC_ARM and IAR?

RonEld:
@andresag01 The FEATURE_NOT_SUPPORTED support is niuce to have, but I will work on it once ARMmbed/mbedtls#1716 is merged, so this isn't a blocker for this PR

benchmark/main.cpp
andresag01: Thanks for clarifying. I have added the checks for the AES feature unavailable only.

benchmark/main.cpp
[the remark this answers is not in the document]
andresag01: I tried one version of GCC_ARM only. However, I do not expect the code to be too different with the others. After all, if a feature is not enabled, its corresponding function will be empty and there is only one assignment and a return in there. Then in main there is only a call to that function and a check which map to the instructions above (or something similar to that).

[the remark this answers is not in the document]
RonEld: I agree there shouldn't be much difference, personally, I prefer not to have a call to empty functions, but that's a matter of style, it's not a blocker for me

andresag01:
@RonEld: I believe I have addressed all (or at least most of) your comments.

andresag01:
I have tested the changes in K64F with GCC_ARM and it seems to work fine. However, when I try it with ARMCC I get the following errors:
```
mbedtls_hmac_drbg_seed() returned -0x5180
  !  MD - Failed to allocate memory
mbedtls_pk_parse_key() returned -0x1180
  !  PEM - Failed to allocate memory
mbedtls_mpi_read_string() returned -0x0010
  !  BIGNUM - Memory allocation failed
mbedtls_ecdsa_genkey() returned -0xFFFFFFFF
  !  UNKNOWN ERROR CODE (0001)
mbedtls_ecdh_make_public() returned -0x0010
  !  BIGNUM - Memory allocation failed
mbedtls_ecp_group_load() returned -0x0010
  !  BIGNUM - Memory allocation failed
```
It now seems that we are running out of heap, but I am not very sure why it succeeds in GCC_ARM if the size of data + bss is larger than with ARMCC, so I would expect the heap space to be smaller. I will take a look at that as soon as I get the change.

benchmark/main.cpp
```diff
@@ -56,979 +50,1350 @@
...
+            mbedtls_printf("Failed to compose title string using "
+                           "mbedtls_snprintf(): %d\n", ret);
+            goto exit;
+        }
+
+        memset(buf, 0, sizeof(buf));
+        memset(tmp, 0, sizeof(tmp));
+
+        ret = mbedtls_ccm_setkey(&ccm, MBEDTLS_CIPHER_ID_AES, tmp, keysize);
+        if (ret != 0) {
```
RonEld: CCM could also return `MBEDTLS_AES_FEATURE_NOT_SUPPORTED` in case no alternative CCM

benchmark/main.cpp
```diff
@@ -56,979 +50,1350 @@
...
+        memset(buf, 0, sizeof(buf));
+        memset(tmp, 0, sizeof(tmp));
+
+        cipher_info = mbedtls_cipher_info_from_type(cipher_type);
+        if (cipher_info == NULL) {
+            mbedtls_printf("mbedtls_cipher_info_from_type() returned NULL\n");
+            goto exit;
+        }
+
+        BENCHMARK_FUNC_CALL(title,
```
RonEld: CMAC could also return `MBEDTLS_AES_FEATURE_NOT_SUPPORTED` in case no alternative CMAC.
This is problematic because of the MACRO though

andresag01:
@RonEld: Please note that this fix is already included in this PR

andresag01:
The failures currently reported by the Ci are unrelated to this change. There seems to be an issue with the CI

andresag01:
@RonEld: I added the extra AES unavailable checks to CCM, CMAC and CTR_DRBG

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
 #else
-#define PRINT_ERROR                                    \
-        mbedtls_printf( "FAILED: -0x%04x\r\n", -ret );
-#endif
+#define ecp_clear_precomputed( g )
+#endif /* MBEDTLS_ECP_C */
+
+static unsigned char buf[BUFSIZE];
+static unsigned char tmp[200];
+static char title[TITLE_LEN];
```
Patater: This is a lot of static memory use. Why not use heap or stack instead?

Do we need `tmp` and `buf` at the same time, or could the users of what's currently `buf` use `tmp` instead?

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
+    if (rng_state != NULL) {
         rng_state  = NULL;
+    }
 
-    while( len > 0 )
-    {
+    while (len > 0) {
         use_len = len;
-        if( use_len > sizeof(int) )
+        if(use_len > sizeof(int)) {
```
Patater: Add a space after `if`.

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
-            mbedtls_ecp_point_free( &grp->T[i] );
-        mbedtls_free( grp->T );
-    }
-    grp->T = NULL;
-    grp->T_size = 0;
-}
-#else
-#define ecp_clear_precomputed( g )
-#endif
+MBED_NOINLINE static int benchmark_arc4() {
```
Patater: Why "NOINLINE"? If you allow the compiler to inline this, and it ends up as an empty function (by configuration choices), there shouldn't be any effect (no stub function made that does nothing).

benchmark/main.cpp
```diff
@@ -156,7 +156,11 @@ do {                                                                        \
                                                                             \
     for (i = 1, alarmed = 0, t.attach(alarm, 1.0); !alarmed; i++)           \
     {                                                                       \
-        if ((ret = (CODE)) != 0) {                                          \
+        ret = CODE;                                                         \
+        if (ret == MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE) {                   \
```
RonEld: Have you tested this with changing the AES code to return `MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE` on 256 bit key? I think this will cause the benchmark to stop instead of skipping the test.

In addition, in future Unavailable errors, this will probably not work when we don't loop and overwrite the ret in next call, but this is not part of this PR

benchmark/main.cpp
```diff
@@ -519,6 +527,8 @@ MBED_NOINLINE static int benchmark_aes_gcm()
 
         ret = mbedtls_gcm_setkey(&gcm, MBEDTLS_CIPHER_ID_AES, tmp, keysize);
         if (ret == MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE) {
+            /* Do not consider this as a failure */
```
RonEld: Note: Obviously this will work only for AES feature unavailable, and not in case an alternative gcm implementation returns the uncreated yet `MBEDTLS_GCM_FEATURE_UNAVAILABLE`

This is not a review comment, but an notice for future implementation

benchmark/main.cpp
```diff
@@ -560,7 +570,11 @@ MBED_NOINLINE static int benchmark_aes_ccm()
         memset(tmp, 0, sizeof(tmp));
 
         ret = mbedtls_ccm_setkey(&ccm, MBEDTLS_CIPHER_ID_AES, tmp, keysize);
-        if (ret != 0) {
+        if (ret == MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE) {
```
RonEld: Note: Obviously this will work only for AES feature unavailable, and not in case an alternative ccm implementation returns the uncreated yet `MBEDTLS_CCM_FEATURE_UNAVAILABLE`

This is not a review comment, but an notice for future implementation

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
+            mbedtls_printf("mbedtls_ecdh_make_public() returned -0x%04X\n",
+                           -ret);
+            goto exit;
+        }
 
-        if( mbedtls_ecp_group_load( &ecdh.grp, MBEDTLS_ECP_DP_CURVE25519 ) != 0 ||
-            mbedtls_ecdh_gen_public( &ecdh.grp, &ecdh.d, &ecdh.Qp, myrand, NULL ) != 0 )
-        {
-            return( 1 );
+        ret =mbedtls_ecp_copy( &ecdh.Qp, &ecdh.Q );
```
Patater: This isn't Mbed OS style...

benchmark/main.cpp
```diff
@@ -747,6 +747,7 @@ MBED_NOINLINE static int benchmark_ctr_drbg()
 {
     int ret = 0;
     const char *nopr_title = "CTR_DRBG (NOPR)";
+    const char *pr_title = "CTR_DRBG (PR)";
```
RonEld: why not use the existing `title` buffer, and just use the snprintf to write to it the relevant title?

benchmark/main.cpp
```diff
@@ -755,6 +756,7 @@ MBED_NOINLINE static int benchmark_ctr_drbg()
     if (ret == MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE) {
```
RonEld: MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE is not a recoverable error, for ctr_drbg, so I don't think it requires special treatment for this error

benchmark/main.cpp
```diff
@@ -56,979 +50,1198 @@
...
-    "des3, des, aes_cmac, des3_cmac, aes_cbc, \r\n"                       \
-    "aes_ctr, aes_gcm, aes_ccm,\r\n"                                      \
-    "havege, ctr_drbg, hmac_drbg,\r\n"                                    \
-    "rsa, dhm, ecdsa, ecdh.\r\n"
-
-#if defined(MBEDTLS_ERROR_C)
-#define PRINT_ERROR                                            \
-        mbedtls_strerror( ret, ( char * )tmp, sizeof( tmp ) ); \
-        mbedtls_printf( "FAILED: %s\r\n", tmp );
+#define BENCHMARK_FUNC_CALL(TITLE, CODE)                                    \
```
Patater: For additional code savings, and easier maintainability (no giant macros), this could be made into a static function that accepts a function pointer and a string or two. This would mean wrapping a function call you want to benchmark with another function that doesn't take any parameters.

benchmark/main.cpp
RonEld: `tmp` and `buf` are used at the same time. `buf` is the buffer of length 1024 that is the input \ output of the current operation.
`tmp` is an additional buffer used for the current operation, such as key, iv, signature etc..
I think, however, that the size of `tmp` should be decreased significantly. I am not sure htat more than 32 bytes is really used

benchmark/main.cpp
RonEld: I agree, and perhaps adding a list of the optional valid return codes that the function can return, such as the different FEATURE_UNAVAILABLE errors ( TBD for the implementation of the return codes)

benchmark/main.cpp
andresag01: I think this review comment is based on an old commit. Please note that later on there is a commit that fixes these style problems by running astyle with the Mbed OS configuration file.

benchmark/main.cpp
andresag01: I think this review comment is based on an old commit. Please note that later on there is a commit that fixes these style problems by running astyle with the Mbed OS configuration file.

benchmark/main.cpp
andresag01: The NOINLINE is because without it the compiler will inline functions into main() with high optimization levels and dramatically increase the stack size causing the application to fail...

andresag01: In a later commit I changed location of the preprocessor checks to surround the entire function and the call in main, so there will not be an empty function added when the feature is disabled.

Note that @RonEld had a similar comment, so I changed this. As I discussed with him, I prefer the other approach because it decreases the number of preprocessor checks with only negligible effect in code size and runtime. Please find below my answer to @RonEld's original comment:

>I think code size here is not really a problem. When the feature is not defined as a macro then the compiler (GCC) will generate the following code for an empty function:
```
0000181c <benchmark_havege()>:
    181c:	2000      	movs	r0, #0
    181e:	4770      	bx	lr
```
>Then there will be the following call in the main:
```
    36aa:	f7fe f8b7 	bl	181c <benchmark_havege()>
    36ae:	b9b8      	cbnz	r0, 36e0 <main+0x37c>
```
>Also, I checked and when multiple functions are empty, then the compiler will reuse one of the empty functions instead of emitting multiple ones to reduce code size. In any case, the overhead is 2.5 words (10 bytes) which take about 8-10 clock cycles to execute in a Cortex-M0...

benchmark/main.cpp
andresag01: >Have you tested this with changing the AES code to return MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE on 256 bit key? I think this will cause the benchmark to stop instead of skipping the test.

I did test it and it seems to work.

>In addition, in future Unavailable errors, this will probably not work when we don't loop and overwrite the ret in next call, but this is not part of this PR

Not sure what you mean. Please bear in mind that this piece of code is inside a while(0) loop. The break statement exits that while(0) (which is in the macro and will always be there. I dont think it really matters whether the whole macro itself is used within a loop or not.

benchmark/main.cpp
andresag01: I am not sure what you mean. The special treatment is only so that the console printout shows up nicely in the terminal. Perhaps there is something I am missing?...

benchmark/main.cpp
andresag01: I am not sure what is the point of using title and snprintf(). There is nothing to format here, these two variables are just a short hand to not have to write the string twice in the function body.

benchmark/main.cpp
andresag01: I have expanded the documentation in the code to explain that the checks need to be extended as more FEATURE_UNAVAILABLE errors are added to the library.

benchmark/main.cpp
andresag01: > This is a lot of static memory use. Why not use heap or stack instead?
I would like to avoid putting these into the stack because that would increase the likely of stack overflows, which is one of the motivations for this PR. These are currently as globals because there are shared by all functions, and in any case they were allocated statically in the original benchmark, so I just left it like that. I could move it to the heap if you feel strongly about it, I don't mind either way...

> Do we need tmp and buf at the same time, or could the users of what's currently buf use tmp instead?
As @RonEld explained, we do need both buf and tmp. I will try to reduce the buffer sizes though.

andresag01:
@RonEld @k-stachowiak @sbutcher-arm @Patater: I had another look at the memory problems we are having with this application. I fixed a few minor problems regarding error checking and then I decided to investigate the out-of-memory errors a bit more. I ran the application in a desktop computer and found that there was actually a memory corruption issue, which I presume was causing the problems we were having. After these fixes, benchmark passes in K64F with ARMCC. I will run the full CI once more to see if there are failures in other platforms/toolchains.

benchmark/main.cpp
andresag01: @Patater, @RonEld: I understand your concerns, but there were a lot of other issues to fix with this application. While code size is important, the main problems I had to look into was the RAM problems. To save time I decided to not do something like what you are suggesting and preserve the main structure of each test case, but inside separate functions.

How strongly about this? Should we just leave it like this for now?

andresag01:
I believe I have addresses or replied to everyones comments. Let me know what are your thoughts.

benchmark/main.cpp
RonEld: what I mean is, for AES256, for example:
Let's say that AES256_CBC will return the `MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE `, so the set key will succeed, but the `mbedtls_aes_crypt_cbc` will return the error code. `ret` is now the error `MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE ` , and we finish the loop in [line 435](https://github.com/ARMmbed/mbed-os-example-tls/pull/194/files#diff-11d19f91de60e62a1952ab3e8b07cd0cR435) causing the function `benchmark_aes_cbc()` to return an error code, changing the `exit_code` to `MBEDTLS_EXIT_FAILURE`. We will not exit the test, but I wouldn't consider this a failure.

As for other modules, which will have a future NOT SUPPORTED error, it's basically the same, the specific benchmark test function will return an error code, even if we don't really consider a FEATURE UNAVAILABLE error a real error code

benchmark/main.cpp
RonEld: I think that in all the other `MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE` we should print and skip the specific test, as it's a reasonable error to return.
For ctr_drbg, it means that ctr_drbg will fail, as it uses AES256 only, unless configured to 128 bit, as introduced in https://github.com/ARMmbed/mbedtls/pull/1902
So, this error is not a valid error for ctr_drbg, IMHO.
If it's only for the printout, well, the `PRINT_ERROR` will handle this, no?

benchmark/main.cpp
RonEld: Have you tried reducing the `tmp` buffer size? From my brief look, it is used for small buffers, such as keys and ivs.
I think that 150 bytes can also be reduced. Couldn't 64 bytes be enough?

andresag01:
@RonEld: I do not think I can reduce the tmp buffer size much further as it is also used to hold ECDSA signatures. If you run the program in x86 with ASAN and then start reducing the buffer size you will see that the smallest size it does not fail with is 139 or so (if I remember correctly). I just upper bounded it to 150...

benchmark/main.cpp
andresag01: I still do not understand what the change request is here. We are already printing out the test and the string "Feature Unavailable" when AES is not available. This is equivalent to skip because we are not letting the benchmark application failed for that. I do not see the point of printing the actual error string corresponding to MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE and this would be inconsistent with the checks in other parts of this application

RonEld:
@andresag01 Thanks for addressing my question on the `tmp` buffer

Please also address my earlier comments, on the FEATURE UNAVAILABLE causing the test to fail. This will probably cause a CI failure, if the CI also checks for the return code, and not only the log

benchmark/main.cpp
RonEld: Because, when we skip a test, it's a valid error, and the return code of the benchmark should be success.
On ctr_drbg, on the other hand, when `MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE ` is returned, it is a real error, and we should fail the test, and developers should be aware that they will need to modify thei ctr_drbg to use 128 bit key size(after https://github.com/ARMmbed/mbedtls/pull/1902 is merged)

andresag01: Ok, I see what you mean, but I do not agree with it. If ctr_drbg does not run because it has no AES that is not a bug or a real failure according to what we have done in the benchmark application. Its just that the primitive is unavailable under the current configuration. Isnt this comparable to saying that AES-256 is not available so its an actual error because the user should change to something like AES-128?

RonEld: No, because for ctr_drbg, the usage of AES256 is embedded within the algorithm and not configurable, and not under the user's discretion.
After it will be configurable to 128, I still think that it should be an error, and not just skipped, because it's an algorithm using AES, and not just a primitive AES, but I won't insist on it, after it will be configurable

mazimkhan:
retest

benchmark/main.cpp
Patater: Is this still a problem?

benchmark/main.cpp
Patater: It's not configurable to use AES-128, yet. After it is configurable, we can make this an error. Is any further action needed here?

benchmark/main.cpp
andresag01: Unfortunately, the macro does make it a bit awkward to add the extra checks. I did add [this](https://github.com/ARMmbed/mbed-os-example-tls/pull/194/files#diff-11d19f91de60e62a1952ab3e8b07cd0cR160) line of code to check for MBEDTLS_AES_FEATURE_NOT_SUPPORTED on calls like this.

I agree that the code is not ideal, but my main motivation for this refactor was to patch up the reliability issues. So I tried to minimize the development time by no changing all these macros. Perhaps we should consider replacing them when it is needed to check more feature not supported errors.

benchmark/main.cpp
andresag01: Unfortunately the code in the latest version of this PR is not really comparable with the original version of the application because a lot more code has been added to improve reliability (for example, all the error checks). However, I am fairly certain that if we removed this, the code would have increased because there are a lot more function prologues and epilogues, but I expect the amount of code regarding these macros to be relatively similar.

andresag01:
@k-stachowiak @RonEld @Patater: I think I have answered everyone's review comments. Please take a look and let me know if there is any more rework that you would like to do before merging. Also, it seems that the example is only failing on KW24D with IAR due to dynamic memory allocation problems. I suspect that this is plainly because we are either having fragmentation problems or the heap is too small in that target. Do we need to investigate this further? Does anyone have access to that target?

benchmark/main.cpp
RonEld: Correct, this is not configurable. This is why It is a failure error, that should cause the CI to fail, and we should not have special treatment for this specific error, in case of ctr drbg

RonEld:
@andresag01 
https://github.com/ARMmbed/mbed-os-example-tls/pull/194#discussion_r215153651 was not addressed yet

In addition, in case of ctr drbg, I still think that *because* it is not configurable to have an AES 128 key, receiving the error `MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE ` is not a valid error, and we should not ignore it.

benchmark/main.cpp
andresag01: Ok, I see what you mean. I will fix that.

andresag01:
@RonEld:
* Thanks for the explanation. I will fix [#194 (comment)](https://github.com/ARMmbed/mbed-os-example-tls/pull/194#discussion_r215153651).
* With regards to the CTR_DRBG issue, I do not agree with it because its not like the target is broken, just not implemented/unavailable. However, you are the one maintaining and developing the error codes, so I will go along and change to the behaviour as you asked.

andresag01:
retest

andresag01:
@Patater @k-stachowiak @RonEld: I have reworked the PR. Could you please review?

benchmark/main.cpp
```diff
@@ -260,6 +260,8 @@ MBED_NOINLINE static int benchmark_md4()
 
     BENCHMARK_FUNC_CALL("MD4", mbedtls_md4_ret(buf, BUFSIZE, tmp));
 
+    ret = 0;
```
k-stachowiak: While this code is correct, I find it confusing. I would consider setting the `ret`'s value inside the benchmark macro - it would express exactly what the commit's title says: "if feature is unavailable, return 0".

benchmark/main.cpp
```diff
@@ -795,12 +795,7 @@ MBED_NOINLINE static int benchmark_ctr_drbg()
     mbedtls_ctr_drbg_init(&ctr_drbg);
 
     ret = mbedtls_ctr_drbg_seed(&ctr_drbg, myrand, NULL, NULL, 0);
-    if (ret == MBEDTLS_ERR_AES_FEATURE_UNAVAILABLE) {
-        /* Do not consider this as a failure */
-        mbedtls_printf(HEADER_FORMAT "Feature unavailable\n", nopr_title);
-        mbedtls_printf(HEADER_FORMAT "Feature unavailable\n", pr_title);
-        goto exit;
-    } else if (ret != 0) {
+    if (ret != 0) {
```
k-stachowiak: I don't mind it handled this way, but for the consistency's sake I'd consider leaving the alternative, having both branches fail, but print different error messages (if that makes sense from the functional standpoint).

andresag01: I see what you mean, but IMHO this is already implemented. Note that PRINT_ERROR actually prints the output of mbedtls_strerror() so there will be a different error message that clearly states whether FEATURE_UNAVAILABLE is the cause of the problem. I think it is redundant to further split this.

andresag01:
@k-stachowiak: I think [this](https://github.com/ARMmbed/mbed-os-example-tls/pull/194#discussion_r219515716) regarding `ret = 0` is the last comment to resolve. I thought this would be easy to follow because the program will reach the `ret = 0` statement only IF everything succeeded. In case of failure, control would jump straight to the `exit` label. Please note that this is the approach often used in Mbed TLS itself, so I thought this was acceptable. What do you think?

k-stachowiak:
@andresag01 I recall similar solutions in Mbed TLS, but I think the macro names are different, e.g. starting with `CHECK_`. This one says `BENCHMARK_`, which doesn't imply any control flow semantics, and yet it can branch. However, since other reviewers did not challenge this solution, and I think it has taken more time than it is worth (as I mentioned, I find the code correct), I will change the status of my review.

andresag01:
@RonEld @k-stachowiak: I looked into the out-of-memory failure and I think the problem is that IAR forces the heap size to be set at compile time and this size is not adjusted dynamically depending on stack usage, etc (as with the allocators used in GCC and ARMCC).

The heap for the KW24D target is set to just 16K, but the target has about 64K and we are only using about 9K in static data. So there should be plenty of memory to run the application! (I think!). I raised [this](https://github.com/ARMmbed/mbed-os/issues/8306) ticket in Mbed OS to ask whether there is a mechanism to configure this in the build system or something to increase the heap size without manually changing the linker script. Currently waiting for a reply...

@Patater @RonEld @k-stachowiak @sbutcher-arm: I do not think this problem is related to the application. Do you think this is a blocker for merging?

andresag01:
Rebased the changes to sync with latest merged to master.

sbutcher-arm:
@RonEld - Can you please re-review? We need to get this merged as the fixes will restore us to working on some broken targets.

andresag01:
@sbutcher-arm: This PR already has 2 approvals, so removing the 'needs review' tag.

sbutcher-arm:
retest

sbutcher-arm:
So this PR seems to fix the tests previously failing on the following targets:
```
build-benchmark-K64F-ARM
build-benchmark-NUCLEO_F429ZI-ARM
build-benchmark-KW24D-ARM
```
The top two were memory problems with ECDSA, and the final one `build-benchmark-KW24D-ARM` failed with ECDHE.

However, we still have failing targets:
```
build-tls-client-UBLOX_EVK_ODIN_W2-IAR
build-tls-client-UBLOX_EVK_ODIN_W2-GCC_ARM
build-tls-client-UBLOX_EVK_ODIN_W2-ARM
build-tls-client-NUCLEO_F429ZI-GCC_ARM
build-benchmark-KW24D-IAR
```

The KW24D just doesn't have enough memory for the example, so we'll need to work out a way of excluding that test. The others are network problems, and I suspect `build-tls-client-NUCLEO_F429ZI-GCC_ARM` was just a transient error, because the test is fragile to network problems.

So this PR does fix what it said it did and is good to merge, but unfortunately reveals that some failing tests were hiding other failing tests. So that means it's good to merge - but we have more bugs to fix on some targets. I'll raise those separately.

sbutcher-arm:
retest

sbutcher-arm:
Retesting to see if the issue with `build-tls-client-NUCLEO_F429ZI-GCC_ARM` goes away on a second go.

PR #198, opened 2018-08-31
Don't use easy-connect, use NetworkInterface::get_default_instance()

SeppoTakalo:
After Mbed OS 5.9, there is API to as a default network interface from Mbed OS.

See: https://os.mbed.com/docs/v5.9/reference/configuration-connectivity.html#selecting-the-default-network-interface

During Mbed OS 5.10 all driver should begin to support this, including external drivers.

Also, as some drivers are entering Mbed OS repository, our examples cannot anymore refer to easy-connect which would pull in those exact same drivers. Leads to duplicated code.

This is needed to get this working with 5.10

### DIFF
--- a/tls-client/HelloHttpsClient.cpp
+++ b/tls-client/HelloHttpsClient.cpp
@@ -21,8 +21,6 @@
 
 #include "HelloHttpsClient.h"
 
-#include "easy-connect.h"
-
 #include "mbedtls/platform.h"
 #include "mbedtls/config.h"
 #include "mbedtls/ssl.h"
@@ -33,6 +31,7 @@
 
 #include <stdint.h>
 #include <string.h>
+#include "mbed.h"
 
 const char *HelloHttpsClient::DRBG_PERSONALIZED_STR =
                                                 "Mbed TLS helloword client";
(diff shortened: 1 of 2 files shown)


tls-client/HelloHttpsClient.cpp
```diff
@@ -232,20 +231,16 @@ int HelloHttpsClient::configureTCPSocket()
...
+    NetworkInterface *network = NetworkInterface::get_default_instance();
     if(network == NULL) {
-        mbedtls_printf("easy_connect() returned NULL\n"
-                       "Failed to connect to the network\n");
+        mbedtls_printf("ERROR: No network interface found!\n");
         return -1;
     }
+    ret = network->connect();
+    if (ret != 0) {
+        printf("Error! network->connect() returned: %d\n", r);
```
RonEld: compilation error: r->ret

change to  `mbedtls_printf()`

RonEld:
Since using the NetworkInterface is needed for 5.10, but prior we need to use easy-connect, we should have some run-time flag \ version tag to decide whether we use NetworkInterface or easy-connect.
It is not unlikely that one will use an older version of Mbed OS on latest examples branch.

SeppoTakalo:
NetworkInterface API is already published in 5.9, so this should work on at least Ethernet enabled boards.
In 5.10 timeframe all other driver types should become supported.

So you do not necessary need flagging.

RonEld:
Yes, I agree, I was thinking of a more backwards compatible to older Mbed OS version, but for that, we have the specific version tags that should be used.

Patater:
Examples get tagged for the release they go with. Do we really need a configuration flag?

RonEld:
>Examples get tagged for the release they go with. Do we really need a configuration flag?

No, I agree that the tag is enough. But we should consider merging #193 prior to this PR, and create tag for 5.9.x  before, and then merge this PR, and create tag for 5.10

0xc0170:
> This is needed to get this working with 5.10

Please treat this with priority. 5.10 released and this example should be updated

SeppoTakalo:
Fixed the errors pointed out in this review.
Rebased on top of master.

SeppoTakalo:
Now the application should work on WiFi boards as well.

There is example `mbed_app.json` in the Socket example documentation
https://github.com/ARMmbed/mbed-os-example-sockets#selecting-the-network-interface

PR #199, opened 2018-09-03
Change mbedtls_platform_context parameter to NULL

RonEld:
`mbedtls_platform_setup()` and `mbedtls_platform_teardown()` now ignore the context parameter, since https://github.com/ARMmbed/mbed-os/pull/7099 was merged. Changing the sent parameter to NULL, to avoid misleading.

### DIFF
--- a/authcrypt/main.cpp
+++ b/authcrypt/main.cpp
@@ -24,15 +24,14 @@
 #include "mbedtls/platform.h"
 
 int main() {
-    mbedtls_platform_context platform_ctx;
     int exit_code = MBEDTLS_EXIT_FAILURE;
 
-    if((exit_code = mbedtls_platform_setup(&platform_ctx)) != 0) {
+    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
         printf("Platform initialization failed with error %d\n", exit_code);
         return MBEDTLS_EXIT_FAILURE;
     }
 
-    Authcrypt *authcrypt = new Authcrypt(&platform_ctx);
+    Authcrypt *authcrypt = new Authcrypt();
 
     if ((exit_code = authcrypt->run()) != 0) {
         mbedtls_printf("Example failed with error %d\n", exit_code);
@@ -41,6 +40,6 @@ int main() {
 
     delete authcrypt;
 
-    mbedtls_platform_teardown(&platform_ctx);
+    mbedtls_platform_teardown(NULL);
     return exit_code;
 }
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -343,7 +343,6 @@ exit:
 }
 #endif /* MBEDTLS_SHA512_C */
 
-
 #if defined(MBEDTLS_ARC4_C)
 MBED_NOINLINE static int benchmark_arc4()
 {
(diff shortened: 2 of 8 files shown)


authcrypt/main.cpp
```diff
@@ -24,15 +24,14 @@
 #include "mbedtls/platform.h"
 
 int main() {
-    mbedtls_platform_context platform_ctx;
     int exit_code = MBEDTLS_EXIT_FAILURE;
 
-    if((exit_code = mbedtls_platform_setup(&platform_ctx)) != 0) {
+    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
```
sbutcher-arm: This is assigning an error code from `mbedtls_platform_setup()` to an `exit_code` used to exit `main()`. They are two different sets of values - and the error values don't map to valid exit codes. Initialise `exit_code` to `MBEDTLS_EXIT_SUCCESS` and remove these follow on assignments. They're not needed.

benchmark/main.cpp
```diff
@@ -1015,20 +999,19 @@ static int benchmark( int argc, char *argv[], mbedtls_platform_context* ctx )
 }
 
 int main(void) {
-    mbedtls_platform_context platform_ctx;
     int exit_code = MBEDTLS_EXIT_FAILURE;
 
-    if((exit_code = mbedtls_platform_setup(&platform_ctx)) != 0) {
+    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
```
sbutcher-arm: Again, I think you should remove the assignment to `exit_code`, and initialise it to `MBEDTLS_EXIT_SUCCESS` instead.

hashing/main.cpp
```diff
@@ -157,20 +152,19 @@ static int example(mbedtls_platform_context* ctx)
 }
 
 int main() {
-    mbedtls_platform_context platform_ctx;
     int exit_code = MBEDTLS_EXIT_FAILURE;
 
-    if((exit_code = mbedtls_platform_setup(&platform_ctx)) != 0) {
+    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
```
sbutcher-arm: Same review comment as before - remove the assignment to`exit_code`, and initialise it to `MBEDTLS_EXIT_SUCCESS` instead.

tls-client/main.cpp
```diff
@@ -50,10 +50,9 @@ const int SERVER_PORT = 443;
  */
 int main()
 {
-    mbedtls_platform_context platform_ctx;
     int exit_code = MBEDTLS_EXIT_FAILURE;
 
-    if((exit_code = mbedtls_platform_setup(&platform_ctx)) != 0) {
+    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
```
sbutcher-arm: Same comment as above - remove the assignment to `exit_code`.

authcrypt/main.cpp
RonEld: @sbutcher-arm I disagree.
Although the exit code is not the return code of the functions, we use it for logging the error code. When we exit the application upon failure, the return code will always be `MBEDTLS_EXIT_FAILURE`

RonEld:
@sbutcher-arm I answered your comment.
I think it is good to assign the `exit_code` with the error codes, as it is helpful in the logging of the errors. 
Upon failure, the applications return `MBEDTLS_EXIT_FAILURE`

authcrypt/main.cpp
sbutcher-arm: Well, I disagree. ;) I think it's reusing the variable for two different purposes with sets of values that don't necessarily mix. But it's a pre-existing problem, so I think I can ignore it for the moment, as the code works, just doesn't match my expectations.

sbutcher-arm:
@RonEld - Can you please rebase this and resolve the conflicts?

authcrypt/main.cpp
RonEld: Sure, I created #205 to track this

RonEld:
@sbutcher-arm I have rebased to resolve the conflicts

sbutcher-arm:
retest

sbutcher-arm:
@RonEld - I hate to ask - but can you please rebase?

RonEld:
@sbutcher-arm Done. Please review

sbutcher-arm:
retest

sbutcher-arm:
CI is failing, but only with known issues (of which we have far too many!), so this can be merged.

Issue #210, opened 2018-10-31
Benchmark example times out on KW24D, on ARM compiler


k-stachowiak:
The example times out while performing benchmark for `DHE-2048`. Surrounding log:

```
[1540134513.96][CONN][RXD]   RSA-4096                 :   20077 ms/private
[1540134527.98][HTST][INF] test suite run finished after 600.77 sec...
[1540134527.98][HTST][ERR] Expected output [\s+DHE-2048\s*:\s*\d+ ms/handshake] not received in log.
```

k-stachowiak:
Note that it is similar to #208

RonEld:
@k-stachowiak Does the example timeout when setting `MBEDTLS_MPI_MAX_SIZE` to 512 ?

k-stachowiak:
@sbutcher-arm @Patater Based on #231 I conclude, that the problem doesn't occur anymore. I suggest closing this issue.

ciarmcom:
@k-stachowiak thank you for raising this issue.Please take a look at the following comments:

Could you add some more detail to the description? A good description should be at least 25 words.
What target(s) are you using?
What toolchain(s) are you using?
What version of Mbed OS are you using (tag or sha)?
It would help if you could also specify the versions of any tools you are using?
How can we reproduce your issue?

NOTE: If there are fields which are not applicable then please just add 'n/a' or 'None'.This indicates to us that at least all the fields have been considered.
Please update the issue header with the missing information, the issue will not be mirroredto our internal defect tracking system or investigated until this has been fully resolved.

ciarmcom:
@k-stachowiak it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@k-stachowiak it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@k-stachowiak it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@k-stachowiak it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@k-stachowiak it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

k-stachowiak:
I still think we can close it. Let me free some @ciarmcom 's cycles and close it.

Issue #208, opened 2018-10-23
Benchmark example times out on UBLOX EVK ODIN W2, on GCC_ARM


k-stachowiak:
The example times out while performing benchmark for `dh-3072`. Surrounding log:

    [CONN][RXD]   DHE-3072                 :    8571 ms/handshake
    [HTST][INF] test suite run finished after 600.36 sec...
    [HTST][ERR] Expected output [\s+DH-3072\s*:\s*\d+ ms/handshake] not received in log.

ciarmcom:
ARM Internal Ref: IOTSSL-2611

k-stachowiak:
Note that it is similar to #210

sbutcher-arm:
GCC_ARM is now passing, but IAR is failing at the same point, as can be seen in the log for build [#816](https://jenkins-internal.mbed.com/blue/organizations/jenkins/mbedtls-release/detail/mbedtls-release/816/pipeline/94/).

RonEld:
@k-stachowiak I don't have a board to test, but could you check if setting `MBEDTLS_MPI_MAX_SIZE` to 512 fixes this issue? I have noticed that the default size of 1024 bytes causes stack overflow in many cases.

k-stachowiak:
@sbutcher-arm This issue seems to be fixed by #213, but that has not been approved yet. If it is an acceptable approach to remove the tests, then the issue may be considered resolved (even though this is a workaround, rather than a fix).

@RonEld I don't have the board either, but I raised a test PR (#219) to see the result in the CI run.

RonEld:
@k-stachowiak Thank you for running the test PR.

If reducing the MPI size to 512 will work, I will make a PR changing the default MPI size

RonEld:
@k-stachowiak Yes, I noticed, thanks
The lack of memory probably needs further investigation, but at least it fixes the timeout

k-stachowiak:
@sbutcher-arm @Patater Based on #231 I conclude, that the problem doesn't occur anymore. I suggest closing this issue.

Issue #138, opened 2017-12-14
Memory allocation failure in benchmark example IAR + KW24D


mazimkhan:
Error:
```
RSA-4096                 :  FAILED: RSA - The private key operation failed : BIGNUM - Memory allocation failed
```

IAR 7.8
KW24D
CI log https://jenkins-internal.mbed.com/job/mbedtls-mbed-os-examples-pr/75/execution/node/592/log/
mbed-os 5.7

@MarceloSalazar please check.

MarceloSalazar:
@maclobdell would you like to review this for the KW24D platform?

maclobdell:
@bentcooke  - did you recently make a heap size change to get KW24D working in another example?  Perhaps that could be relevant to this issue.  Please share any details you have.

bentcooke:
@maclobdell  I had to reduce the heap size (20k->15k) and move the nanostack heap into the system heap (`"mbed-mesh-api.use-malloc-for-heap": true`,) in order to get the mesh minimal example working on the KW24D.

ciarmcom:
ARM Internal Ref: IOTSSL-2427

k-stachowiak:
The impact of this issue has increased by one more test failing to allocate memory on the heap. Currently there are two allocation failures in the BIGNUM module:

    ...
    RSA - The private key operation failed : BIGNUM - Memory allocation failed
    ...
    DHM - Making of the public value failed : BIGNUM - Memory allocation failed
    ...

sbutcher-arm:
I understand the underlying issue is there's not enough heap available on this target for the test, which has been raised as issue [#8306](https://github.com/ARMmbed/mbed-os/issues/8306).

We've got no way of externally determining if the target can support the test or not, so I'd like to propose we change the benchmark to only test `RSA-2048`, and skip `RSA-4096`. `RSA-4096` is demanding in memory and in my view it's better to disable the test until [#8306](https://github.com/ARMmbed/mbed-os/issues/8306) is fixed, which is open ended and has no immediate resolution.

Equally, I suggest we disable the Diffie-Hellman tests. They're not part of the default configuration of Mbed TLS in Mbed OS, so why run them? The [lines enabling them in the config.h file](https://github.com/ARMmbed/mbed-os-example-tls/blob/97321fb91997c180ef35512c3f6ee78cabd2b2ae/benchmark/mbedtls_config.h#L92), can just be removed.

andresag01:
[...]

Issue #209, opened 2018-10-23
No network interface found in the CI for UBLOX EVK ODIN W2


k-stachowiak:
Regardless of the compiler used, the following log can be found in the CI reports for the ODIN W2 device:

    [CONN][RXD] Using Mbed OS 5.10.2
    [CONN][RXD] ERROR: No network interface found!
    [CONN][RXD] 
    [CONN][RXD] FAIL
    [HTST][INF] test suite run finished after 600.68 sec...

ciarmcom:
ARM Internal Ref: IOTSSL-2610

k-stachowiak:
@sbutcher-arm @Patater Based on #231 I conclude, that the problem doesn't occur anymore. I suggest closing this issue.

PR #213, opened 2018-11-06
Remove non-default, memory-heavy tests from the benchmark application

k-stachowiak:
This PR removes memory-heavy tests from the benchmark application. They are not used by default and they don't work for all the tested devices due to the memory restrictions (at least ones implied by the build configurations).

Fixes (works around until we have resolved the memory configuration issues) #138 

edit:
It apparently also fixes (works around...) #208 and #210.

### DIFF
--- a/benchmark/mbedtls_config.h
+++ b/benchmark/mbedtls_config.h
@@ -89,10 +89,6 @@
 #define MBEDTLS_RSA_C
 #endif
 
-#if !defined(MBEDTLS_DHM_C)
-#define MBEDTLS_DHM_C
-#endif
-
 #if !defined(MBEDTLS_ECDSA_C)
 #define MBEDTLS_ECDSA_C
 #endif
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -903,7 +903,6 @@ MBED_NOINLINE static int benchmark_rsa()
     mbedtls_rsa_context *rsa;
     const char *rsa_keys[] = {
         RSA_PRIVATE_KEY_2048,
-        RSA_PRIVATE_KEY_4096,
     };
     size_t i;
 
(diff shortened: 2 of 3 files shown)


k-stachowiak:
retest

k-stachowiak:
The only failing tests are due to a known issue: #209

benchmark/mbedtls_config.h
```diff
@@ -89,10 +89,6 @@
 #define MBEDTLS_RSA_C
 #endif
 
-#if !defined(MBEDTLS_DHM_C)
-#define MBEDTLS_DHM_C
-#endif
-
```
andresag01: @k-stachowiak: Thanks for the PR, but in my opinion this is not sufficient. I think you should explicitly use #undef if you want that particular feature to be disabled.

k-stachowiak: @andresag01 Thanks for the review.

This change was made based on the following reasoning: there are some features used here, that are not even enabled by default. Therefore, we should not enable them forcibly. Should, in the future, the features become enabled by default (and break this app), we may return to this issue, and e.g. explicitly disable them.

Looking at this from another perspective, there are probably more features in Mbed TLS that would break the benchmark, but we don't explicitly disable the in this file, since they are not enabled by default. Should `DHM` be treated specially in this case?

Please consider my arguments and confirm, if you think the feature should still be explicitly disabled.

k-stachowiak:
@andresag01 Do you feel satisfied by my reply to your comment? If so, can you approve the PR, or maybe you have more change requests?

sbutcher-arm:
Passes CI with the exception of the known broken tests on UBLOX.

adbridge:
@k-stachowiak  @sbutcher-arm Why has this just been merged to master when the 5.11 release has not been made yet and all fixes should consequently still be going to the 5.11.0-oob branch ??!!

Issue #205, opened 2018-10-10
sample applications should not set exit_code as return statement of functions


RonEld:
As commented in https://github.com/ARMmbed/mbed-os-example-tls/pull/199#discussion_r224024552 
internal functions return code should be set to a variable different than `exit_code`.
It's not a functionality bug, but it shows bad practice

andresag01:
I have created the PR #221 to fix the issue reported here.

ciarmcom:
@RonEld thank you for raising this issue.Please take a look at the following comments:

Could you add some more detail to the description? A good description should be at least 25 words.
What target(s) are you using?
What toolchain(s) are you using?
What version of Mbed OS are you using (tag or sha)?
It would help if you could also specify the versions of any tools you are using?
How can we reproduce your issue?

NOTE: If there are fields which are not applicable then please just add 'n/a' or 'None'.This indicates to us that at least all the fields have been considered.
Please update the issue header with the missing information, the issue will not be mirroredto our internal defect tracking system or investigated until this has been fully resolved.

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
@RonEld it has been 5 days since the last reminder. Could you please update the issue header as previously requested?

ciarmcom:
[...]

PR #221, opened 2018-12-06
Fix unsuitable use of exit_code in authcrypt

andresag01:
The exit_code variable in authcrypt is meant to hold the main() return value to the system. However, it was also being shared with the return value for internal functions. This PR splits the two uses and introduces a variable ret for holding internal return values.

Addresses https://github.com/ARMmbed/mbed-os-example-tls/issues/205


sbutcher-arm:
retest

Issue #26, opened 2016-11-30
tls-client example failing in UBLOX_EVK_ODIN_W2 dev board


andresag01:
The tls-client compiles correctly with ARMCC and gcc for the UBLOX_EVK_ODIN_W2 target using mbed-os-5.3.0-rc1. However, no output is observed in the serial terminal when the application is run.

To reproduce, compile the program, flash and then connect the serial terminal:
```
mbed compile -t GCC_ARM -m UBLOX_EVK_ODIN_W2
```

ciarmcom:
ARM Internal Ref: IOTSSL-1108

andreaslarssonublox:
I tested this and it seems to work for me if the EMAC macro is removed from the device_has in targets.json for UBLOX_EVK_ODIN_W2. If this is not removed I get the same behavior as you have with no output.
The default MAC is now Wi-Fi introduced in this commit:
https://github.com/ARMmbed/mbed-os/commit/20b7f057214a9edcda554c50ebec99e51f9756ed

bulislaw:
Shouldn't TLS work on both Ethernet and WiFi?

bulislaw:
Ok, synced with Martin -> the issue is by default Ethernet is not enabled on Odin board, with EMAC macro enabled the WiFi is active. The example doesn't support WiFi, therefore we have a fail.

yanesca:
It has run successfully when compiled in the online IDE.

andresag01:
@0xc0170: Removing `EMAC` from the `device_has` for the target in the targets.json fixes the problem. However, is there a way to force the target to use Ethernet instead of Wifi without modifying mbed OS's targets.json to remove the `EMAC`?

bulislaw:
You can use `device_has_remove` in mbed_app.json like in https://github.com/ARMmbed/mbed-os-example-sockets/blob/master/mbed_app.json

andresag01:
@bulislaw: Thanks for the information. I have created a PR with the suggested change. https://github.com/ARMmbed/mbed-os-example-tls/pull/41

andresag01:
The fix for this issue has been merge...

PR #222, opened 2018-12-06
Set default network driver on Ublox target to ethernet

dgreen-arm:
The tls-client example needs updating for the UBLOX_EVK_ODIN_W2 platform to correctly set the default network interface to be ethernet.

The previous fix was discussed in #26, however the suggested fix in https://github.com/ARMmbed/mbed-os-example-tls/issues/26#issuecomment-275036468 now shows the change we wish to make here.

### DIFF
--- a/tls-client/mbed_app.json
+++ b/tls-client/mbed_app.json
@@ -31,7 +31,7 @@
              "platform.stdio-convert-newlines": true
         },
         "UBLOX_EVK_ODIN_W2": {
-            "target.device_has_remove": ["EMAC"]
+            "target.network-default-interface-type" : "ETHERNET"
         }
     }
 }


sbutcher-arm:
retest

sbutcher-arm:
CI passes for gcc but fails on IAR and Arm C Compiler. Can you please look into it @dgreen-arm?

dgreen-arm:
retest

dgreen-arm:
retest

dgreen-arm:
retest

dgreen-arm:
It looks like the failures were due to infrastructure issues, all the tests have successfully run in the last two test runs.

Patater:
> This change is making an implicit preference for ethernet over Wifi on a specific platform.

I believe we were already trying to make that preference before this PR. The issue is that how one elects to prefer Ethernet over WiFi has changed, and this PR updates it for the (currently) correct way.

Maybe we don't need to pick Ethernet anymore? I'd expect our examples to work fine on WiFi, and CI can do WiFi.

dgreen-arm:
Preference for ethernet for the Ublox platform is specifically done as its the only target that we currently test that has the default interface being wifi, all the other targets have ethernet as their default interface.

sbutcher-arm:
@patater - Because we test this sample with a variety of boards in the CI, the boards we test with are always configured for ethernet.

To fix this board in CI, we're making it explicit. That was my point.

So do we want to document this? I guess there's a bunch of mbed boards we're also not configured for.

Patater:
I believe we are making it explicit already for all targets our example runs on via this line to select Ethernet as the interface to use: https://github.com/ARMmbed/mbed-os-example-tls/blob/master/tls-client/mbed_app.json#L9

It's probably worth documenting in the README that the example requires Ethernet. We had issues in the past on the Odin board because a selection of Ethernet meant WiFi at some point in the past (2016), as we didn't have WiFi credentials present. It looks like Odin understands Ethernet means Ethernet these days, and perhaps the workaround isn't necessary anymore.

dgreen-arm:
The README for the TLS client does state that ethernet is required.

I've tried removing the lines changing the default driver, the Odin board still fails as the example tries to get the default network interface, which is Wifi. There's two ways to fix this I know of: changing the default driver to ethernet, or changing the example to specifically get an ethernet interface.

Given that the TLS example currently requires ethernet, which would be a better solution? Change the example to specifically get an ethernet interface, or set the default driver to ethernet for all boards?

kjbracey-arm:
If you always want this to be Ethernet, then you can unconditionally set that default to `ETHERNET`, regardless of target.

Or you could change the code to ask for `EthInterface::get_default_instance()`, not `NetworkInterface::get_default_instance()`.

I would suggest that in an example it's better for the default mbed_app.json to use the target default network interface.  Maybe you could have a separate mbed_app.json for CI that forces `ETHERNET` for all targets - I believe various CI jobs do select alternative JSON files.

PR #232, opened 2019-01-22
Allow the use of PSA Crypto in TLS client example

hanno-arm:
__Summary:__ This PR allows the use of PSA Crypto in the TLS client example by adding the necessary `psa_crypto_init()` call to the initialization code.

To use, define `MBEDTLS_USE_PSA_CRYPTO` in `mbedtls_entropy_config.h` before invoking `mbed compile`. No further changes are necessary.

### DIFF
--- a/tls-client/main.cpp
+++ b/tls-client/main.cpp
@@ -35,6 +35,9 @@
 #include "mbed.h"
 
 #include "mbedtls/platform.h"
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#include "psa/crypto.h"
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
 #include "HelloHttpsClient.h"
 
@@ -51,6 +54,24 @@ const int SERVER_PORT = 443;
 int main()
 {
     int exit_code = MBEDTLS_EXIT_FAILURE;
+
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    /*
+     * Initialize underlying PSA Crypto implementation.
+     * Even if the HTTPS client doesn't make use of
+     * PSA-specific API, for example for setting opaque PSKs
+     * or opaque private keys, Mbed TLS will use PSA
+     * for public and symmetric key operations as well as
+     * hashing.
+     */
+    psa_status_t status;
+    status = psa_crypto_init();
+    if( status != PSA_SUCCESS )
+    {
+        printf("psa_crypto_init() failed with %d\r\n", status );
+        return MBEDTLS_EXIT_FAILURE;
+    }
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
         printf("Platform initialization failed with error %d\r\n", exit_code);
--- a/tls-client/main.cpp
+++ b/tls-client/main.cpp
@@ -55,6 +55,11 @@ int main()
 {
     int exit_code = MBEDTLS_EXIT_FAILURE;
 
+    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
+        printf("Platform initialization failed with error %d\r\n", exit_code);
+        return MBEDTLS_EXIT_FAILURE;
+    }
+
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     /*
      * Initialize underlying PSA Crypto implementation.
@@ -73,10 +78,6 @@ int main()
     }
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
-    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
-        printf("Platform initialization failed with error %d\r\n", exit_code);
-        return MBEDTLS_EXIT_FAILURE;
-    }
     /*
      * The default 9600 bps is too slow to print full TLS debug info and could
      * cause the other party to time out.
--- a/tls-client/mbedtls_entropy_config.h
+++ b/tls-client/mbedtls_entropy_config.h
@@ -36,4 +36,14 @@
 #undef MBEDTLS_MPI_MAX_SIZE
 #define MBEDTLS_MPI_MAX_SIZE        256
 
+/* This macro determines whether Mbed TLS uses its own legacy crypto library
+ * or an implementation of the PSA Crypto API such as Mbed Crypto.
+ *
+ * To confirm the use of PSA Crypto, you may enable debugging by setting
+ * HELLO_HTTPS_CLIENT_DEBUG_LEVEL in HelloHttpsClient.h and look for
+ * PSA-related debugging output on the serial line.
+ *
+ * Uncomment this to use the PSA Crypto API. */
+//#define MBEDTLS_USE_PSA_CRYPTO
+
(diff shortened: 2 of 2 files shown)


tls-client/main.cpp
```diff
@@ -52,6 +55,24 @@ int main()
...
+    /*
+     * Initialize underlying PSA Crypto implementation.
+     * Even if the HTTPS client doesn't make use of
+     * PSA-specific API, for example for setting opaque PSKs
+     * or opaque private keys, Mbed TLS will use PSA
+     * for public and symmetric key operations as well as
+     * hashing.
+     */
+    psa_status_t status;
+    status = psa_crypto_init();
```
RonEld: I think this should come after `mbedtls_platform_setup()`

hanno-arm: @RonEld Why?

RonEld: Perhaps the best location of calling `psa_init()` is within [`mbedtls_platform_setup()` in `mbed_platform_setup()`](https://github.com/ARMmbed/mbed-os/blob/master/features/mbedtls/platform/src/platform_alt.c#L37)

hanno-arm: > because mbedtls_platform_setup() initializes the hw acceleration engine, and it psa_init() is dependent on the driver,

@RonEld If the underlying implementation of PSA Crypto uses Mbed TLS somehow, then it is responsible for setting it up properly; after all, we might be developing an application which just uses PSA Crypto, and not Mbed TLS, or the PSA Crypto implementation might be entirely independent of Mbed TLS. Hence, `psa_crypto_init()` must work on its own. 

Still, I guess it must be considered on a per-platform basis what the chosen PSA Crypto implementation's `psa_crypto_init()` does, to avoid conflicts with other hardware initialization procedures such as `mbedtls_platform_setup()`; e.g., `mbedtls_platform_setup()` might be defined to already include `psa_crypto_init()`. However, given the generic nature of both `psa_crypto_init()` and `mbedtls_platform_setup()`, I don't think one can make a decision on that in the generality of this example. 

What about adding a comment saying something like "Depending on the implementation of `psa_crypto_init()` and `mbedtls_platform_setup()` for a particular platform, a particular order of these calls might be necessary, or one of them might subsume the other."

Ping @gilles-peskine-arm

RonEld: One comment, in Mbed-OS, `mbedtls_platform_setup()` is already implemented.

RonEld: @gilles-peskine-arm  I agree that one should call the other, but we should take into consideration platforms that don't support yet `MBEDTLS_USE_PSA_CRYPTO`
If `psa_crypto_init()` will call `mbedtls_platform_setup()`, but psa is not defined, then the platform initialization will not happen.
However, if `mbedtls_platform_setup()` will call `psa_crypto_init()` in case `MBEDTLS_USE_PSA_CRYPTO` is defined, there won't be a functional problem, except a bad example(), so it is better to keep the `psa_crypto_init()` in the example.
However, I think it should be *after* the call to `mbedtls_platform_setup()` which initializes the platform ( not only hw acceleration), and `psa_crypto_init()` initializes the crypto module only. In real practice they basically do the same. I believe that `psa_crypto_init()` will eventually replace the other.
Perhaps we should do the following, to set good example and to support non PSA boards in the transition period:
Have `psa_ceypto_init()` call `mbedtls_platform_setup()`, and in the example:
~~~
#if defined(MBEDTLS_USE_PSA_CRYPTO)
    /*
     * Initialize underlying PSA Crypto implementation.
     * Even if the HTTPS client doesn't make use of
     * PSA-specific API, for example for setting opaque PSKs
     * or opaque private keys, Mbed TLS will use PSA
     * for public and symmetric key operations as well as
     * hashing.
     */
    psa_status_t status;
    status = psa_crypto_init();
    if( status != PSA_SUCCESS )
    {
        printf("psa_crypto_init() failed with %d\r\n", status );
        return MBEDTLS_EXIT_FAILURE;
    }
#else

    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
        printf("Platform initialization failed with error %d\r\n", exit_code);
        return MBEDTLS_EXIT_FAILURE;
    }
#endif /* MBEDTLS_USE_PSA_CRYPTO */
~~~

gilles-peskine-arm: One solution could be to have `mbedtls_platform_setup` and `psa_crypto_init` call each other (`mbedtls_platform_setup` would only call `psa_crypto_init` under `MBEDTLS_USE_PSA_CRYPTO`), and to make `psa_crypto_init` reentrant. But that would make the implementation `psa_crypto_init` fragile in multithreaded environments, where a concurrent call to `psa_crypto_init` must not return immediately but rather wait until the first call has completed.

RonEld: I think that in *mbed-os*, we should have `psa_crypto_init()` call `mbedtls_platform_setup()`. In Mbed OS, we have a reference counter within the platform context, and a second call to `mbedtls_platform_setup()` will not do anything except increase the counter.

hanno-arm: I think the scope of `psa_crypto_init()` in comparison to `mbedtls_platform_init()` should be discussed somewhere else. Until it is settled, I think the PR should go in as-is, because as it stands, `mbedtls_platform_init()` as used in Mbed OS does not initialize PSA.

RonEld: I agree that the scope should be discussed somewhere else.
i still think that `psa_crypto_init()` should come *after* the call to `mbedtls_platform_setup()`, as the latter initializes the platform, while the former initializes the crypto module, which *might* need resources initialized by `mbedtls_platform_setup()`

hanno-arm: @RonEld I disagree with that - there might be _cross_ dependency between the two calls (to be discussed), but it cannot be that `psa_crypto_init()` _needs_ `mbedtls_platform_init()`, because it's a standalone library and you could use it without Mbed TLS.

RonEld: Yes, it's a standalone library.
However, `mbedtls_platform_setup()` was first and foremost intended to initialize the HW acceleration engine. `psa_crypto_init()` may need that engine.

hanno-arm: @RonEld Sorry, I could only repeat what I just said. There cannot be any dependency of `psa_crypto_init()` on Mbed TLS, in particular on `mbedtls_platform_setup()`.

hanno-arm:
@gilles-peskine-arm @RonEld Please see the latest commit.

tls-client/main.cpp
hanno-arm: @RonEld Omitting `mbedtls_platform_setup()` if `MBEDTLS_USE_PSA_CRYPTO` is defined doesn't work because the function might perform other platform initialization tasks not related to PSA crypto. Are you ok letting this rest until https://github.com/ARMmbed/mbed-crypto/issues/24 is settled?

RonEld: At the moment, there are no other platform initialization tasks done in Mbed OS. The first intent for the platform initialization functions were to initialize hw acceleration. In the future it may be considered for other initialization (assuming it won't get deprecated), after the psa crypto will be fully integrated.
I will approve for the time being

sbutcher-arm:
Reviewed, approved, and passing CI, so can be merged.

PR #257, opened 2019-08-15
Pass an integer value for callback interval as API has been updated

hugueskamba:
Mbed OS PR for API change: https://github.com/ARMmbed/mbed-os/pull/11236

### DIFF
--- a/benchmark/main.cpp
+++ b/benchmark/main.cpp
@@ -154,7 +154,7 @@ do {                                                                        \
     mbedtls_printf(HEADER_FORMAT, TITLE);                                   \
     fflush(stdout);                                                         \
                                                                             \
-    for (i = 1, alarmed = 0, t.attach(alarm, (s_timestamp_t)1); !alarmed; i++) {           \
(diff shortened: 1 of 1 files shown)


benchmark/main.cpp
```diff
@@ -154,7 +154,7 @@ do {                                                                        \
     mbedtls_printf(HEADER_FORMAT, TITLE);                                   \
     fflush(stdout);                                                         \
                                                                             \
-    for (i = 1, alarmed = 0, t.attach(alarm, 1.0); !alarmed; i++) {         \
+    for (i = 1, alarmed = 0, t.attach_s(alarm, 1); !alarmed; i++) {         \
```
evedon: This depends on https://github.com/ARMmbed/mbed-os/pull/11236 being released

hugueskamba: That's the whole point of the PR and it says so in the description.

hugueskamba:
Closing this PR as the solution in ARMmbed/mbed-os#11236 requires no change from the API users.

PR #303, opened 2021-07-19
CMake: Make applications the top level CMake projects

rwalton-arm:
So CMake can work correctly we need to define our project before we add
dependencies, otherwise the project we depend on will be registered as
the current project.


Patater (approved):

Patater:
Travis passed, but failed to report back

